Issue #86, opened 2024-07-24
Flatpak package and publishing on Flathub


k8ieone:
Really liking the app so far, it would be nice if the project was available as a Flatpak.

taylorhadden:
Sounds like a good idea to me! I'm not really a linux user myself. Is this something you're knowledgeable on / would mind helping out with?

k8ieone:
No, infortunately I haven't done anything like this before. But I'll try looking into it, my guess is that it shouldn't be too difficult.

k8ieone:
Alright, I did some initial research and I'm trying to throw a manifest together. I have a question tho. Would you like to publish the finished application on Flathub yourself?

Flathub's policy is that users are allowed to submit applications that they're not the developer of. https://docs.flathub.org/docs/for-app-authors/submission#theres-an-app-that-id-like-to-see-on-flathub-but-im-not-the-developer

But it seems that having the developer publish the application is preferred. https://docs.flathub.org/docs/for-app-authors/submission#someone-else-has-put-my-app-on-flathubwhat-do-i-do

This is still in the future, the manifest is nowhere near ready yet. I just wanted to shoot this question out beforehand.

I'm hoping this might not even matter because once a repository in https://github.com/flathub has been created the publisher gets write permissions and I hope more people can be invited to also have write access. This way we could both manage the application on Flathub. https://docs.flathub.org/docs/for-app-authors/submission/#after-submission

taylorhadden:
I have no strong preferences about being the primary publisher. If you or anybody else want to help out here, I'd be quite pleased! That's why Tangent is open source now.

k8ieone:
I might have spoken too soon :sweat_smile:

Tangent being an Electron app using `npm` for builds complicates things because Flathub doesn't allow network access from the sandbox where the application is built. The workaround for this is to pull the dependencies before the build process starts and then use `npm` in offline mode for the build. They have [tooling](https://github.com/flatpak/flatpak-builder-tools/tree/master/node) and a [guide](https://docs.flatpak.org/en/latest/electron.html) for this, but the Typewriter fork being a local module seems to break their script for dependency fetching. I'm afraid I'm not enough of an `npm` guru for this :sweat_smile:

In case anyone wants to pick this up, feel free to do so.

I'll post an update if I manage to get any further, but my hopes aren't so high anymore :pensive: 

This is what I have so far:

```yaml
id: io.github.suchnsuch.Tangent
runtime: org.freedesktop.Platform
[...]

PR #92, opened 2024-08-27
Update `package-lock.json` and switch submodule to HTTP

k8ieone:
In order to get Flatpak builds working, I've had to make a few changes.

1. Update `package-lock.json`
2. Switch the Typewriter submodule to HTTP

I've had to switch from SSH because Github doesn't allow pulling from SSH unless you add the machine's SSH key to a Github account. Since the builds will run on Flathub's CI servers, this is not feasible.

In case you need write access to the submodule from the main repo, I recommend using `insteadOf` in your git config.

Example:

```ini
[url "ssh://git@github.com"]
	insteadOf = https://github.com
```

Related to #86

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "lib/typewriter"]
 	path = lib/typewriter
-	url = git@github.com:taylorhadden/typewriter.git
+	url = https://github.com/taylorhadden/typewriter.git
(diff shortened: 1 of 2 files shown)


taylorhadden:
Good stuff!

That other package will need to be resolved at some point somehow. It's not exactly important, but it's part of the monorepo and will come back next time the lock file gets updated.

taylorhadden (approved):

PR #108, opened 2024-11-02
Add appstream metainfo and .desktop file

k8ieone:
Related to #86, Flathub maintainers would like the metainfo and .desktop files to be upstreamed.


taylorhadden:
Thanks for putting this forward!

Are the `.github` elements of the filenames necessary for flathub? In theory, there's no github attachment. What does the `io` part refer to?

Also, could these be moved to a "flatpak" subfolder?

k8ieone:
We can name the files whatever we want in this repo. Moving them elsewhere is also not an issue.

To explain the current name - it follows the application ID. The application ID spec is here https://docs.flathub.org/docs/for-app-authors/requirements#control-over-domain-or-repository

taylorhadden:
Okay, cool! If they don't mind how they're named right now, I don't care.

Thanks for moving them!

taylorhadden (approved):

PR #118, opened 2024-12-28
Update metainfo file with new versions

k8ieone:


taylorhadden (approved):

taylorhadden:
Hmm, I should have probably merged that off of 3a0895796. Right now this is sitting on top of a bunch of unstable v0.9.x stuff.

k8ieone:
Oh yea, that would be better 😅 

We'll keep it in mind for future releases. You can also update the metainfo file yourself if you remember.

It shouldn't matter for now, you'll probably do a release for switching to Electron 34.

PR #123, opened 2025-01-18
Bump Electron to 34.0.0, regenerate package-lock

k8ieone:
Related to #86

https://github.com/electron/electron/issues/43819 indeed seems to have been fixed. I was able to use the Tangent Flatpak on systems where the file picker was previously broken.

I also did a few basic tests (opening the app, loading my workspace) to make sure the new Electron version doesn't break anything major, but you might want to check yourself.

### DIFF
--- a/apps/tangent-electron/electron-builder.json
+++ b/apps/tangent-electron/electron-builder.json
@@ -4,7 +4,7 @@
 	"copyright": "Copyright © 2022 Taylor Hadden",
 
 	"generateUpdatesFilesForAllChannels": true,
-	"electronVersion": "33.0.2",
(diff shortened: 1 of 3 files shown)


taylorhadden:
Great, thank you! Do you also need to update the flatpak files?

k8ieone:
Oh yea, I forgot about that. The metainfo file needs a new version with the release notes.

Can you prepare that please? I won't be home today.

taylorhadden:
I've implemented and released this as v0.8.4 and v0.9.0-alpha.6.

k8ieone:
Awesome, thanks

PR #139, opened 2025-02-18
Creation rules with note content templates

shawngraham:
Hi - I worked out a way to set the content of new notes from workspace-settings.json and via the creation rules editor. Slight modifications to default/workspace-settings.json to create a template key, and to creationRule.ts, createNewFile.ts, creationRuleEditor.svelte to read the key and populate the note and editor as required. It may also have required some fiddling of either main/File.ts or main/Files.ts.
Screenshot 2025-02-17 at 3 27 46 PM

Please excuse my clunkiness.


shawngraham:
ack, I left a bunch of crap in the Documentation folder; apologies. Meant to delete that. Sorry, not very handy with git.

Documentation/.tangent/workspace-settings.json
taylorhadden, line 1: Looks like the tab indentation was swapped to spaces in this file. Please be sure to keep the same spacing formats.

apps/tangent-electron/defaults/workspace-settings.json
taylorhadden, line 1: Looks like this has the same tabs-to-spaces indentation issue.

I have some additional thoughts:

1. Since there is a `nameTemplate` already, we probably want to call the new key `contentTemplate` or something similar.
2. I don't want to include any content templates for the default "Inbox" and "Daily Note" rules. I want to retain Tangent's minimalism here. I don't expect everybody to use content templates.
3. Looks like the "Project Log" is DH-specific. We should keep these defaults very stripped down.

apps/tangent-electron/src/app/model/nodeViewStates/FeedViewState.ts
```diff
@@ -9,7 +9,7 @@ import FeedView from 'app/views/node-views/FeedView.svelte'
 import FeedSettingsView from 'app/views/node-views/FeedSettingsView.svelte'
 import type ViewStateContext from './ViewStateContext'
 import { NoteViewState } from '.'
-import { getNode, indexOfMatch, TreeNodeOrReference } from 'common/nodeReferences'
```
taylorhadden, line 12: Not sure why this is commented out. Surely this causes compilation issues?

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -248,7 +248,16 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
 
 				this.workspace.navigateTo(nav)
 			}
-	
+            if (newNode instanceof File && context?.rule) {
+                const templateValue = typeof context.rule.template === 'string' ?
```
taylorhadden, line 252: Doesn't the template need to have its replacement variables injected here?

Also, this block is using spaces for indentation instead of tabs.

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -290,4 +299,4 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
 		}
 		return 'Creates a new note'
 	}
-}
```
taylorhadden, line 293: This change can be dropped.

apps/tangent-electron/src/app/views/creation-rules/CreationRuleEditor.svelte
```diff
@@ -79,6 +80,10 @@ function templateDependencies(template) {
 			</figcaption>
 		</figure>
 	</details>
+    <label title="Defines the content of the new note using markdown.">
```
taylorhadden, line 83: More spaces for indentation.

apps/tangent-electron/src/app/views/creation-rules/CreationRuleEditor.svelte
```diff
@@ -154,4 +159,21 @@ figcaption {
 	}
 }
 
+textarea { /* Add style for textarea */
```
taylorhadden, line 162: How many of these rules are actually required to get the desired effect? We may want to provide a default setup in input.scss.

The comments here are not really necessary.

Also, spaces instead of tabs.

apps/tangent-electron/src/common/settings/CreationRule.ts
```diff
@@ -106,40 +107,36 @@ const creationStoreOptions: ObjectStoreOptions = {
...
+            if (!name) return 'New Creation Rule'
+            return name
+        })
+        this.nameTemplate = new WritableStore('%name%')
+        this.folder = new Setting(folderStoreDefinition)
+        this.mode = new Setting(creationModeDefinition)
+        this.description = new Setting(descriptionDefinition)
+        this.showInMenu = new Setting(showInMenuDefinition)
+        this.openInContext = new Setting(openInContextDefinition)
+        this.template = new WritableStore('%template%')  // Add this line
```
taylorhadden, line 133: It looks like this entire file was reformatted (with spaces) in order to add a single line. Please refrain from this.

The default of `'%template%'` seems strange. Surely the default should be null?

package.json
```diff
@@ -18,6 +18,8 @@
 		"apps/tangent-test-workspace-generator"
 	],
 	"devDependencies": {
-		"vitest": "^2.1.8"
+		"vitest": "^2.1.8",
```
taylorhadden, line 21: Did you install new packages here? Why is webpack added to the root?

packages/tangent-html-to-markdown/tsconfig.build.tsbuildinfo
taylorhadden, line 1: Hmm, I likely need to modify the configuration for these packages to not include them...

I'd like to not include them as part of this patch.

test-files/QueryTestWorkspace/.tangent/workspace-settings.json
taylorhadden, line 1: This file and the screenshot workspace file probably don't need to be modified here.

taylorhadden (changes requested):
I'm super please you've taken the initiative here. Content templating is something that the has come up a number of times.

I think the biggest distraction here is with the auto-formatting and the use of spaces for indentation instead of tabs. If we can resolve those, then we can talk about the meat and gristle of the feature.

If you have want please feel free to DM me!

taylorhadden:
I'll chalk it up to a learning experience! If you want to close this and start with another PR, no worries!

PR #150, opened 2025-03-27
Update Electron version, refresh `package-lock.json`

k8ieone:
Can you please update the `tangent-v0.9.0-alpha.9` tag afterwards?


taylorhadden (approved):
I clearly need to pull this into my release routines so you don't have to keep doing this. Thank you!

PR #211, opened 2025-08-01
Refactor: events.ts

justbyitself:
Just a little refactoring. Lesser lines, cleaner code.


justbyitself:
I know it's just a micro refactoring. I like reading clean code. Hopefully, more functional, less imperative. I'd like to know if you are open to this kind of refactoring (and more tests). Of course, it's up to you.
Thank you @taylorhadden.

taylorhadden:
FWIW, I would say the function is exactly as "functional" as it was before. There's no meaningful difference between a one-line ternary and an expanded `if/else`.

If you're up for it, I'd push you more towards improvements and additions to functionality! More tests are definitely good as well (though 100% code coverage is a non-goal).

PR #212, opened 2025-08-02
Chore: update tangent-html-to-markdown devDependencies

justbyitself:
This update let me launch tests with TypeScript 5.


taylorhadden:
This is a good callout. This package hasn't been touched in a while. I actually want to move away from using jest due to all of the issues it has with standard js modules. I'll make that happen.

taylorhadden:
I converted the tests to use `vitest` in f3231954f.

PR #214, opened 2025-08-03
Remove jest from tangent-html-to-markdown and tangent-html-to-markdown

justbyitself:
Removing jest from tangent-html-to-markdown and tangent-html-to-markdown, I can run tests with no problems.


taylorhadden (approved):
Good catch. Missed this in my initial pass.

PR #218, opened 2025-08-07
Refactor directoryView.ts

justbyitself:
Remove unused imports.
Remove redundant string type check.


apps/tangent-electron/src/app/model/directoryView.ts
```diff
@@ -35,13 +34,11 @@ function encodeSortMode({ key, order }: SortMode): string {
 }
 
 export function decodeSortMode(value: string): SortMode {
-	if (typeof value === 'string') {
```
taylorhadden, line 38: There was definitely a reason I did this. I wonder if it was some legacy information. In any case, it was 4 years ago. This change is fine now.

taylorhadden (approved):

PR #364, opened 2026-05-21
Make middle dot [·] and bullet [•] characters also start lists

visrosa:
I can use `\u2022` and `\u00B7` if you'd rather not have unicode in the source.

### DIFF
--- a/apps/tangent-electron/src/common/markdownModel/list.ts
+++ b/apps/tangent-electron/src/common/markdownModel/list.ts
@@ -12,7 +12,7 @@ import NoteParser from './NoteParser'
  * 	7: upper-case single letters
  * 8: Checkbox on top of list
  */
-export const listMatcher = /^([ \t]*)(([-+\*])|((\d+)|([a-z])|([A-Z]))[\.\)])( \[[ x\-]?\])? /
+export const listMatcher = /^([ \t]*)(([-+\*•·])|((\d+)|([a-z])|([A-Z]))[\.\)])( \[[ x\-]?\])? /
 
 /**
  * A subset of the above match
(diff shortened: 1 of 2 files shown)


taylorhadden:
For regex, I would definitely using the unicode escape codes!

Could you also make it so that `•` is a "large" list item? See line 115 of `list.ts`

visrosa:
The new commit adds extensive Unicode character support.

I also added tests but I couldn't actually verify them, `npm run test` gave me errors about playwright and jest and I haven't investigated that fully yet. Kept the commits separate so the tests can be easily rejected if needed.

I should note that the new `standardUnorderedGlyphs` do conflict with the "Change List Glyph on Indent" option, which will turn any of them into a minus-hyphen.

apps/tangent-electron/src/common/markdownModel/list.ts
```diff
@@ -1,5 +1,35 @@
...
+ * U+25e6 ◦	(White Bullet)
+ * U+2619 ☙	(Reversed Rotated Floral Heart Bullet)
+ * U+2767 ❧	(Rotated Floral Heart Bullet)
+ * U+25c9 ◉	(Fisheye (Japanese Bullet)) */
+
+const largeUnorderedGlyphSet = new Set(largeUnorderedGlyphs)
+
+// Build the unordered branch as data so adding glyphs does not disturb the
+// capture-group layout consumed by matchList().
+const unorderedGlyphMatcher = `(?:${[...standardUnorderedGlyphs + largeUnorderedGlyphs]
```
taylorhadden, line 30: I'm curious why you went for a non-capturing group rather than the existing `[]` character set.

taylorhadden (approved):
Tried it out locally. It's very fun to have all these expressive options.

apps/tangent-electron/src/common/markdownModel/list.ts
visrosa: Mostly for readability, the regex got very unwieldy when just adding them all to []

Issue #227, opened 2025-08-23
[issue] RTL languages


hamidb80:
Hi, 

I've just noticed that some components like lists and quotes do not align well with right to left (RTL) languages

<img width="1336" height="1094" alt="Image" src="https://github.com/user-attachments/assets/c7c51991-19ad-4507-a3c9-c07cac431b15" />

hamidb80:
### Problem 7

text direction is left-to-right in card 

<img width="3000" height="2000" alt="Image" src="https://github.com/user-attachments/assets/a2c83c6c-fd61-405a-9f9e-f1a641717f9e" />

taylorhadden:
That last one is odd. There's nothing obviously different. It all has the same "auto".

hamidb80:
> That last one is odd. There's nothing obviously different. It all has the same "auto".

Yeah, so strange

I've tried `text-align: -webkit-auto` on `card` and it seems to solved the problem

<img width="3000" height="1994" alt="Image" src="https://github.com/user-attachments/assets/d2451328-c2dc-46da-817c-0d8725c8ff82" />

hamidb80:
> that's kinda funny. perhaps rtl languages should use something like `+` instead of `x`

> Yeah, I think that's a solid notion. It puts it a bit outside the normal markdown context, but it's a reasonable idea. It's a shame I can't mark the span as "non-language".

Could this be made optional in preferences—like letting users choose whether to tick off TODO list items with `+` or `x`?


after that, I would consider this issue to be solved;
nice work Taylor! thanks a lot

taylorhadden:
The card issue should be fixed in v0.12.2-beta.2.

PR #381, opened 2026-06-08
fixes strange behaviour for RTL where `dir=auto` does not work in card view

hamidb80:
as mentioned in https://github.com/suchnsuch/Tangent/issues/227#issuecomment-4644782274


taylorhadden:
Oh good find. This helped me identify some obsolete css that was causing the problem! Fix soon!

PR #385, opened 2026-06-11
ask for workspace option

### LINKED ISSUE #379: Start up at the workspace menu.
Is there an option to make Tangent show workspace selection menu instead of automatically opening the last closed one?

hamidb80:
fixes #379

### DIFF
--- a/apps/tangent-electron/src/app/App.svelte
+++ b/apps/tangent-electron/src/app/App.svelte
@@ -186,7 +186,6 @@ window.addEventListener('keyup', (event:KeyboardEvent) => {
 })
 
 // Try to load a default workspace
-
 api.getWorkspace()
 .then(state => setState(state.globalSettings.askWorkspaceEveryTime ? null : state))
 .catch(e => {
--- a/apps/tangent-electron/src/app/App.svelte
+++ b/apps/tangent-electron/src/app/App.svelte
@@ -25,6 +25,7 @@ import { updateMermaidStyle } from './style/mermaidStyle'
 // Doing this here so that mhchem is loaded
 import 'katex/contrib/mhchem/mhchem'
 import { setDefaultUseKeyCode } from './utils/shortcuts'
+import type Settings from 'common/settings/Settings';
 
 let applicationState: 'initializing' | 'choosingWorkspace' | 'buildingWorkspace' | 'usingWorkspace' | 'error' = 'initializing'
 let showLoading = false
@@ -187,8 +188,8 @@ window.addEventListener('keyup', (event:KeyboardEvent) => {
 
 
 // startup behaviour
-api.settings.getRaw().then(s => {
-	if (s['startupBehaviourAboutWorkspace'] == 'ask'){
+api.settings.getSafe().then(s => {
+	if ((s as Settings).startupBehaviourAboutWorkspace.value == 'ask'){
 		setState(null)
 	}
 	else {
--- a/apps/tangent-electron/src/app/App.svelte
+++ b/apps/tangent-electron/src/app/App.svelte
@@ -189,7 +189,7 @@ window.addEventListener('keyup', (event:KeyboardEvent) => {
 
 // startup behaviour
 api.settings.getSafe().then(s => {
-	if ((s as Settings).startupBehaviourAboutWorkspace.value == 'ask'){
+	if ((s as Settings).startupWorkspaceBehaviour.value == 'ask'){
 		setState(null)
 	}
 	else {
--- a/apps/tangent-electron/src/app/App.svelte
+++ b/apps/tangent-electron/src/app/App.svelte
@@ -204,7 +204,6 @@ api.settings.get().then(s => {
 	console.log(e)
 })
 
-
 let title = 'Tangent'
 async function setState(state: WorkspaceInitState) {
 	if (showLoadingTimeout) {
--- a/apps/tangent-electron/src/app/App.svelte
+++ b/apps/tangent-electron/src/app/App.svelte
@@ -187,7 +187,7 @@ window.addEventListener('keyup', (event:KeyboardEvent) => {
 
 
 // startup behaviour
-api.settings.get().then(s => {
+api.settings.getRaw().then(s => {
 	if (s['startupBehaviourAboutWorkspace'] == 'ask'){
 		setState(null)
 	}
--- a/apps/tangent-electron/src/app/App.svelte
+++ b/apps/tangent-electron/src/app/App.svelte
@@ -187,13 +187,8 @@ window.addEventListener('keyup', (event:KeyboardEvent) => {
 
 // Try to load a default workspace
 
-api.getWorkspace().then((state) => {
-	if (state.globalSettings.askWorkspaceEveryTime){
-		setState(null)
-	}
-	else
-		setState(state)
-})
+api.getWorkspace()
+.then(state => setState(state.globalSettings.askWorkspaceEveryTime ? null : state))
 .catch(e => {
 	console.error('Could not load default workspace.')
 	console.log(e)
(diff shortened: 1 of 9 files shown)


apps/tangent-electron/src/app/views/System/Database.svelte
```diff
@@ -21,4 +21,7 @@ let settings = workspace.settings
 	<div class="settingsGroup">
 		<SettingView setting={settings.areLinksCaseSensitive} />
 	</div>
+	<div class="settingsGroup">
```
taylorhadden: Let's put this at the top of these settings.

"Database" isn't the _best_ category, but there really isn't a better one. Maybe I'll re-arrange the "Appearance" category as "General" with an "Appearance" header.

apps/tangent-electron/src/common/settings/Settings.ts
```diff
@@ -489,6 +489,12 @@ export default class Settings extends ObjectStore {
 		defaultValue: false
 	})
 
+	askWorkspaceEveryTime = new Setting<boolean>({
```
taylorhadden: I think a better way to present this option would be to have a set of values for what happens on startup.

Maybe the option is called `startupWindows` and the options are called "Last Opened" and "Workspace Selector"?

apps/tangent-electron/src/app/App.svelte
```diff
@@ -186,7 +186,8 @@ window.addEventListener('keyup', (event:KeyboardEvent) => {
 })
 
 // Try to load a default workspace
-api.getWorkspace().then(setState)
+api.getWorkspace()
```
taylorhadden: I don't think we want to make this decision here. A better option would be to choose to not restore previously opened windows on line 205 of `main/index.ts`. You should be able to pull the settings via the `getSettings()` function in `main/settings.ts`.

taylorhadden (changes requested):

hamidb80:
image


demo

hamidb80:
I'm not sure whether `s['startupBehaviourAboutWorkspace']` seems good to you or not

hamidb80:
0. the `getSettings()` function did not work on browser side and must be passed by IPC, since it had functions, ... the IPC could not send it. it had to be converted to raw structure before sending by IPC
1. I had to add the function `getSafeWithHoles()` because I didn't wanted to use `getRawValues()` since it loses the actual data structure that can't be linted by typescript. e.g. `(s as Settings).startupBehaviourAboutWorkspace.value` instead of `s['startupBehaviourAboutWorkspace']`
2. I could easily define new event like `getStartupBehaviour` but that would be too noisy I guess, I tried to use existing `getGlobalSettings` event on IPC that was not used before
3. it wasn't easy as I thought :cry: 

I'm still open to discussion 
thanks in advance

apps/tangent-electron/src/common/settings/Settings.ts
```diff
@@ -489,6 +489,24 @@ export default class Settings extends ObjectStore {
...
 	})
 
+	startupBehaviour = new Setting<'last' | 'ask'>({
+		name: 'Startup Behaviour',
+		description: 'Whether automatically open the last workspace or let you choose one on startup',
+		validValues: [
+			{
+				value: 'last',
+				displayName: 'Last Opened',
+				description: 'Reopen last workspace'
```
taylorhadden: This should be "workspaces" (or "workspace(s)")

apps/tangent-electron/src/app/App.svelte
```diff
@@ -185,10 +186,22 @@ window.addEventListener('keyup', (event:KeyboardEvent) => {
 	}
 })
 
-// Try to load a default workspace
-api.getWorkspace().then(setState)
-.catch(e => {
-	console.error('Could not load default workspace.')
+
+// startup behaviour
+api.settings.getSafe().then(s => {
```
taylorhadden: I think there was a miscommunication about what I was suggesting here. I do not think this file should change. Instead, we should prevent call to re-open the old workspaces in the first place. That happens in the `main/index.ts` file at line 227. That is where the new setting should be checked. With things happening there, we don't need the new IPC path for settings.

taylorhadden (changes requested):

apps/tangent-electron/src/main/index.ts
```diff
@@ -224,7 +224,7 @@ Tangent ${app.getVersion()} Launched With Arguments:`, process.argv)
 				openPathInTangent(openOverride)
 				hasOpenedWindows = true
 			}
-			else if (workspaces.openWorkspaces) {
+			else if (getSettings().startupBehaviour.value == 'last' && workspaces.openWorkspaces) {
```
taylorhadden: 😍

apps/tangent-electron/src/main/settings.ts
```diff
@@ -138,7 +138,7 @@ function postSettings(patch?) {
 // Tests invoke this module outside of an electron context
 if (ipcMain) {
 	ipcMain.handle('getGlobalSettings', async (event) => {
-		return settings
+		return settings // BUG cannot pass `settings` object directly by IPC (since it has functions, ...)
```
taylorhadden: What errors have you seen due to this?

apps/tangent-electron/src/common/settings/Settings.ts
```diff
@@ -489,6 +489,24 @@ export default class Settings extends ObjectStore {
 		defaultValue: false
 	})
 
+	startupBehaviour = new Setting<'last' | 'ask'>({
```
taylorhadden: Can we use the American "Behavior" here (and elsewhere)? Just for consistency.

apps/tangent-electron/src/common/settings/Settings.ts
```diff
@@ -489,6 +489,24 @@ export default class Settings extends ObjectStore {
...
+		description: 'Whether automatically open the last workspace or let you choose one on startup',
+		validValues: [
+			{
+				value: 'last',
+				displayName: 'Last Opened',
+				description: 'Reopen last workspace(s)'
+			},
+			{
+				value: 'ask',
+				displayName: 'Workspace Selector',
```
taylorhadden: I think "Select Workspace" makes sense here. It aligns with the sentence "The startup behavior is to select the workspace" rather than "The startup behavior is to open the workspace selector".

apps/tangent-electron/src/common/settings/Settings.ts
```diff
@@ -489,6 +489,24 @@ export default class Settings extends ObjectStore {
 		defaultValue: false
 	})
 
+	startupBehaviour = new Setting<'last' | 'ask'>({
+		name: 'Startup Behaviour',
+		description: 'Whether automatically open the last workspace or let you choose one on startup',
```
taylorhadden: I don't think we need to describe the options here, as the individual values will do that for us. Perhaps just "What windows Tangent opens on startup."?

Also, a tiny piece: can you add periods to the ends of these descriptions?

taylorhadden (changes requested):

apps/tangent-electron/src/main/settings.ts
hamidb80: The message was not received on the browser side, and no errors were reported.

taylorhadden: Huh, this is actually unused! The initial settings payload is pushed with the workspace when the window is created. Everything after that is patched.

I'll handle that separately.

taylorhadden (changes requested):
This is basically perfect! Just need to remove that settings comment.

taylorhadden (approved):

taylorhadden:
Thank you for the addition!

Issue #367, opened 2026-05-26
add custom commands


hamidb80:
Hi

lots of thanks and appreciation for Tangent!

Would be super useful to be able to add custom commands (e.g. I built a little tool for my language that fixes common typos, and I'd love to hook it into Tangent)

Something like: give it a `name` and a `command` or function that runs:

```
<APP> <PATH_TO_WORKSPACE> <PATH_TO_CURRENT_FILE>
```

Nothing fancy, just passes the current file and workspace path.

Any chance this could be added? 🙏

Thanks again!

hamidb80:
it would be nice if Tangent could provide additional info about the state, like `--cursor:10,17` but that's too much to ask for now

taylorhadden:
Okay, this is a _cool_ idea.

It could be as simple as a string/text box that you can type in a command to. Then you could use replacement tokens like `%WORKSPACE_PATH%` and `%FILEPATH%`.

This would also be super useful in the context of kicking off any number of scripts, including publishing. This seems like a natural next feature to build in the next release cycle.

hamidb80:
yeah I always thought to integrate `git add .`, `git push` into Tangent. with this feature it would be possible + number of arbitrary other things

PR #388, opened 2026-06-16
custom commands

hamidb80:
demo for #367


hamidb80:
# questions
- how to get current workspace directory? I've tried `this.workspace.workspaceFolder.path` but it has `.tangent` suffix
- should I write to global settings ?
- how to get saved cursor position of the current file? (row and column)

hamidb80:
the situation is good, I hit `ctrl+R` (custom set) and then enter the name of the command and it runs

the problem is how to show the output of the command

the easiest and rational thing is to open the console as a separate window (or powershell in windows)

apps/tangent-electron/src/app/modal/ExecCommandDialog.svelte
```diff
@@ -0,0 +1,69 @@
...
+
+function selectOption(option: Script, event) {
+	const vs = workspace.viewState.tangent.getCurrentViewState()
+	const args = [
+		'--file', vs.node.path, 
+		'--workspace', workspace.viewState.directoryView.root.path,
+		
+		...(
+			vs.node.fileType == '.md' 
+			? ["--cursor", (vs as NoteViewState).selection.value.join(',')]
```
hamidb80: @taylorhadden a code prettier! it can capture the selection region and prettifies that

`script --file path/to/file.md --workspace path/to/workspace/ --cursor 100,200`

apps/tangent-electron/src/app/modal/ExecCommandDialog.svelte
```diff
@@ -0,0 +1,69 @@
...
+let workspace = getContext('workspace') as Workspace
+let text: string = ''
+
+type Script = {
+	name: string
+	file: string
+	path: string
+}
+
+export let subject: TreeNode
```
taylorhadden: For new components, let's use the new Svelte 5 Rune syntax.

apps/tangent-electron/src/common/WindowApi.ts
```diff
@@ -104,8 +104,14 @@ export default interface WindowAPI {
 		showInFileBrowser(path: string)
 		/** Open a path in the default format */
 		openPath(path: string)
+
+		findFiles(dir: string, exts: string[]): Promise<string[]>
```
taylorhadden: The renderer's workspace Directory Store tree _should_ know about any `.sh` or `.bat` scripts. You should be able to search that pretty easily instead of adding a new OS path.

apps/tangent-electron/src/main/messages/index.ts
```diff
@@ -19,8 +19,31 @@ import './themes'
 import './urlData'
 import { FileSaveResult } from 'main/File'
 
+import { spawn } from 'child_process'
+
 const log = Logger.get('messages')
 
+
+ipcMain.handle('execCLI', async (event, cmd, args) => {
```
taylorhadden: This handler isn't going to return a `string[]`. Is the intention to return `stdio`?

apps/tangent-electron/src/app/modal/ExecCommandDialog.svelte
taylorhadden: Whenever paths are passed to scripts, we should wrap them in quotes "". This solves the problem of spaces in filepaths.

taylorhadden (changes requested):
There is definitely the beginning of something really cool here. My primary request/recommendation is that we handle this in a very similar way to Creation Rules, where we have a list of objects with many sub-options. This gives the user an opportunity to set a nice display name, define custom shortcuts, and configure other options.

It could also let a user define commands to run inline. Instead of arbitrarily injecting note context into every script call, the user could us wildcards like `$filepath` or `$workspaceDir` when writing an inline script. For example:
```sh
git log -p $filepath
```

I do think that exposing arbitrary scripts is pretty neat. Maybe that is a fallback? If no Tangent Note Action references a script, maybe it shows up as an option?

In either world, I do think that we want all of these actions & scripts to show up in the command palette directly (the same as Creation Rules).

apps/tangent-electron/src/common/WindowApi.ts
hamidb80: yeah I've looked at `directory tree` and it seemed a little strange to me

so I've wrote these messy/hacky to have a demo

apps/tangent-electron/src/main/messages/index.ts
hamidb80: you know I doubt that we should handle commands internally or just give up & open CMD/Console of the OS. I don't want to introduce another VSCode :D

hamidb80: Do you think is it good to have a command control/queue ? We should capture stderr stdout and even keep track of currently running ones

Does it worth it?

I mean it will make Tangent more close to an OS i.e. Tangent Not OS (notes)

taylorhadden: I think doing something with the output is interesting, but I don't think it's critical for a first pass at this.

hamidb80:
in settings

image

image

hamidb80:
apart from handling `stdout` and `stderr`, it's almost ready and working

* UI like creation rules, under `preferences/External Commands`
* works with shortcuts

apps/tangent-electron/src/main/messages/index.ts
hamidb80: I've tried using this but somehow the entire process would crash
```
dialog.showMessageBoxSync(browserWindow, {...})
```

apps/tangent-electron/src/app/views/System/System.svelte
```diff
@@ -17,6 +17,7 @@ import Dictionary from './Dictionary.svelte'
 import Shortcuts from './Shortcuts.svelte'
 import Debug from './Debug.svelte'
 import DocumentationLink from 'app/utils/DocumentationLink.svelte'
+    import ExternalCommands from './ExternalCommandRules.svelte';
```
taylorhadden: Nit: the indentation could be consistent.

apps/tangent-electron/src/common/markdownModel/templates.ts
```diff
@@ -106,6 +106,24 @@ export const dateTemplates: TemplateDefinition[] = [
 	},
 ]
 
+export const commandTemplates: TemplateDefinition[] = [
```
taylorhadden: I don't think these should live here. The templates here are intended for templates that live within a markdown note.

apps/tangent-electron/src/common/settings/ExternalCommand.ts
```diff
@@ -0,0 +1,80 @@
...
+	return (typeof rule.name == 'string' ?  rule.name : rule.name.value) || name
+}
+
+export function willPromptForName(name: string) {
+	return !!name.trim()
+}
+
+const commandTemplateDefinition: SettingDefinition<string> = {
+	name: 'Command Template',
+	description: 'The markdown template file that seeds the initial content of a file.',
```
taylorhadden: The descriptions of these settings have not been updated to reflect their new use for commands.

apps/tangent-electron/src/main/messages/index.ts
```diff
@@ -19,8 +19,30 @@ import './themes'
 import './urlData'
 import { FileSaveResult } from 'main/File'
 
+import { spawn } from 'child_process'
+
 const log = Logger.get('messages')
 
+
+ipcMain.handle('execCLI', async (event, cmd) => {
+	console.log("Executing CLI command: ", cmd)
```
taylorhadden: For logging in the main process, use the appropriate `js-logger` value (otherwise it doesn't get written to the log file). The logger is named `log` in this file.

apps/tangent-electron/src/common/WindowApi.ts
```diff
@@ -23,6 +23,12 @@ export interface UserMessage {
 	message: string
 }
 
+export interface CliStatus {
```
taylorhadden: Perhaps `CliResult` is a better name? This API isn't meant to be polling the process, right? Just give the result?

apps/tangent-electron/src/app/views/external-commands/ExternalCommandRuleTemplateButton.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
```
taylorhadden: Is this actually different from `CreationRuleTemplateButton`? It seems pretty generic. Seems like that component could be used directly instead.

If you want to make it feel more generic, it could be called `TemplateButton` and live in the `app/utils/` directory.

apps/tangent-electron/src/app/views/external-commands/ExternalCommandRuleEditor.svelte
```diff
@@ -0,0 +1,217 @@
+<script lang="ts">
+import { getContext, tick } from 'svelte'
+import { type PathValidationMessages, validatePath } from 'common/trees'
+import { nameFromRule, willPromptForName } from 'common/settings/ExternalCommand'
+import { Workspace } from 'app/model'
+import editable from 'app/utils/editable'
+import SettingView from '../System/SettingView.svelte'
+import { tooltip } from 'app/utils/tooltips'
+import ExternalCommandRuleTemplateButton from './ExternalCommandRuleTemplateButton.svelte'
+    import { commandTemplates } from 'common/markdownModel/templates'
```
taylorhadden: Nitpick: consistent import indentation.

apps/tangent-electron/src/app/views/external-commands/ExternalCommandEditor.svelte
```diff
@@ -0,0 +1,217 @@
...
+
+	exampleNameMessages = messages
+}
+
+/**
+ * Insert the given insertionString into the template at the last-seen selection start position. If the lastSelectionEnd position
+ * is unlike the lastSelectionStart position, then the insertion will be made between the two positions and the selected text will be removed.
+ * @param insertionString the string to insert into the template
+ */
+function insertTextIntoTemplate(
```
taylorhadden, line 26: This looks like a straight copy/paste from `CreationRuleEditor`. Ideally the function would be extracted and shared (though the dependencies required make that a bit of work…)

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { ExternalCommandRuleDefinition } from "common/settings/ExternalCommand"
+import type { NoteViewState } from "../nodeViewStates"
+
+
+export function runExternalCommand(workspace: Workspace, rule: ExternalCommandRuleDefinition){
+	const currentViewState = workspace.viewState.tangent.getCurrentViewState()
+	const ctx = {
+		'file': currentViewState.node.path, 
+		'workspace': workspace.viewState.directoryView.root.path,
+		'cursor': currentViewState.node.fileType == '.md' ? (currentViewState as NoteViewState).selection.value.join(',') : 'nan,nan',
```
taylorhadden: Negative numbers (e.g. `-1`) feel like a reasonable fallback for a "no selection" state. `NaN` is not something that should be used as anything but an error state.

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
```diff
@@ -0,0 +1,75 @@
...
+	workspace.api.os.execCLI(cmd)
+}
+
+interface ExecCliCommandContext extends CommandContext {
+	rule?: ExternalCommandRule
+}
+
+export default class ExecCliCommand extends WorkspaceCommand {
+	constructor(workspace: Workspace) {
+		super(workspace, {group: 'Notes'}) // <---- this makes it available when editing notes
```
taylorhadden: Nitpick: spacing should be `{ group: 'Notes' }`.

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
```diff
@@ -0,0 +1,75 @@
...
+	const shellEscaper = new ShellEscape({shell: 'bash', quote: true}) // TODO change this according to the OS
+	let cmd = rule.commandTemplate.replace(/%([^%]*?)%/g, (match, expr) => shellEscaper.escape(ctx[expr]))
+	workspace.api.os.execCLI(cmd)
+}
+
+interface ExecCliCommandContext extends CommandContext {
+	rule?: ExternalCommandRule
+}
+
+export default class ExecCliCommand extends WorkspaceCommand {
```
taylorhadden: It would be very cool if this command could publish the various pre-defined external commands directly to the command palette. This is how the `CreateNewFileCommand` does it, via `getPaletteActions()`.

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function runExternalCommand(workspace: Workspace, rule: ExternalCommandRuleDefinition){
+	const currentViewState = workspace.viewState.tangent.getCurrentViewState()
+	const ctx = {
+		'file': currentViewState.node.path, 
+		'workspace': workspace.viewState.directoryView.root.path,
+		'cursor': currentViewState.node.fileType == '.md' ? (currentViewState as NoteViewState).selection.value.join(',') : 'nan,nan',
+		'thread': '...'
+	}
+	const shellEscaper = new ShellEscape({shell: 'bash', quote: true}) // TODO change this according to the OS
```
taylorhadden: This `TODO` should definitely be done before this ships.

I also think it makes a lot of sense for the `app`/renderer process to _not_ know a lot about the target shell. The `main` process could be the thing to handle the appropriate shell escapes. It would also be easier to do any introspection about what shells are available.

apps/tangent-electron/src/common/settings/ExternalCommand.ts
```diff
@@ -0,0 +1,80 @@
...
+
+// Used for differentiation in drag/drop
+// TODO: Maybe these could be persistent? Maybe they can help with list sync?
+let externalCommandID = 0
+
+const creationStoreOptions: ObjectStoreOptions = {
+	patchBlockList: ['id']
+}
+
+export default class ExternalCommandRule extends ObjectStore {
```
taylorhadden: The "Rule" name came over from the "Creation Rule". I don't think it's totally appropriate in this context. I think this could just be called `ExternalCommand`. The use of "Rule" can be similarly removed from the rest of the files & variables.

apps/tangent-electron/src/app/modal/ExecCommandDialog.svelte
```diff
@@ -0,0 +1,57 @@
+<script lang="ts">
+import { getContext } from 'svelte'
+import type Workspace from '../model/Workspace'
+import ModalInputSelect from './ModalInputSelect.svelte'
+import type { NoteViewState } from 'app/model/nodeViewStates'
+import type { ExternalCommandRuleDefinition } from 'common/settings/ExternalCommand'
+    import { ShellEscape } from 'app/utils/shell';
```
taylorhadden: Nitpick: indentation consistency.

apps/tangent-electron/src/app/modal/ExecCommandDialog.svelte
```diff
@@ -0,0 +1,57 @@
...
+    import { runExternalCommand } from 'app/model/commands/ExecCliCommand';
+
+let workspace = getContext('workspace') as Workspace
+let text: string = ''
+
+export let commands: ExternalCommandRuleDefinition[]
+
+function filterScripts(c){
```
taylorhadden: This argument should have a proper name and a defined type.

apps/tangent-electron/src/app/modal/ExecCommandDialog.svelte
```diff
@@ -0,0 +1,57 @@
...
+	return commands.filter(s => s.name.includes(c))
```
taylorhadden: The command palette showcases a reasonably robust fuzzy-esque searching system that also features sorting of results and highlighting of what part of a name matches the search. I think that would be definitely something to include here.

apps/tangent-electron/src/app/views/external-commands/ExternalCommandEditor.svelte
```diff
@@ -0,0 +1,217 @@
...
+
+let asksForName = false
+let exampleName = ''
+let exampleNameMessages: PathValidationMessages = []
+
+let templateInput: HTMLInputElement
+
+$: templateDependencies($commandTemplate)
+function templateDependencies(template) {
+	asksForName = willPromptForName(template)
```
taylorhadden: This function doesn't make sense to me in the context of invoking a command. Is it just a remnant of moving this stuff over from the creation rules?

apps/tangent-electron/src/app/views/external-commands/ExternalCommandRuleEditor.svelte
```diff
@@ -0,0 +1,217 @@
...
+		<span>Name Template</span>
+		<input type="text" bind:value={$ruleName} bind:this={templateInput}/>
+	</label>
+	{#if exampleNameMessages?.length}
+		{#each exampleNameMessages as message}
+			<p class={'explanation ' + message.level}>{@html message.message}</p>
+		{/each}
+	{/if}
+	<details>
+		<summary>Name Template Tokens</summary>
```
taylorhadden: This should be called "Command Template Tokens"

apps/tangent-electron/src/app/utils/shell.ts
```diff
@@ -0,0 +1,343 @@
+const supportedShells = ['bash', 'dash', 'busybox', 'zsh', 'csh', 'cmd', 'powershell']
+
+/**
+ * ShellEscape - A class for escaping shell arguments across multiple shells
+ * Extracted from shescape library
```
taylorhadden: Why not use the library directly? Is there a reason to not use it?

taylorhadden (changes requested):
Lots of stuff here. This is _absolutely_ the right direction 🍻.

The largest point of single feedback is that I don't think we want to use "Rule"; we can just call these things "External Commands".

I also want to highlight the query about `shell.ts`. Why not use the referenced library?

apps/tangent-electron/src/app/views/System/System.svelte
hamidb80: I don't know why vscode sometimes goes crazy with automatic imports

apps/tangent-electron/src/app/views/external-commands/ExternalCommandEditor.svelte
hamidb80: Yes, it's just result of copy-pasting

apps/tangent-electron/src/common/markdownModel/templates.ts
hamidb80: OK I'll move to to `common/settings/ExternalCommand.ts`

apps/tangent-electron/src/common/WindowApi.ts
hamidb80: right, it meant to be used just for result

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
hamidb80: Ummm, why this project does not have `.prettierrc` file or something ?

I feel I bother you with indentation and style consistency a lot

apps/tangent-electron/src/app/utils/shell.ts
hamidb80: on second thought, you're right it is not needed

apps/tangent-electron/src/app/views/external-commands/ExternalCommandEditor.svelte
hamidb80: the function is useful but it should live in `utils` or something IMHO

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
hamidb80: I can't think of any. like `git sync` or something?

taylorhadden: Sure! If somebody defined a command, they could then search in the command palette for it by name directly rather than having to jump through two layers.

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
```diff
@@ -2,53 +2,50 @@ import type { Workspace } from ".."
...
+export function runExternalCommand(workspace: Workspace, command: ExternalCommandDefinition){
 	const currentViewState = workspace.viewState.tangent.getCurrentViewState()
-	const ctx = {
+	const context = {
 		'file': currentViewState.node.path, 
 		'workspace': workspace.viewState.directoryView.root.path,
-		'cursor': currentViewState.node.fileType == '.md' ? (currentViewState as NoteViewState).selection.value.join(',') : 'nan,nan',
-		'thread': '...'
+		'cursor': currentViewState.node.fileType == '.md' ? (currentViewState as NoteViewState).selection.value.join(',') : '-1,-1',
+		'thread': '...' // TODO
```
taylorhadden: May as well not include this yet if there's nothing there. You can leave the `TODO: Add thread context`.

apps/tangent-electron/src/common/settings/ExternalCommand.ts
```diff
@@ -1,39 +1,67 @@
...
+
+export interface ExternalCommandDefinition {
 	name: string
 	commandTemplate: string
 	description: string
 }
 
-export function nameFromRule(rule: ExternalCommandRule | ExternalCommandRuleDefinition, name?: string): string {
-	return (typeof rule.name == 'string' ?  rule.name : rule.name.value) || name
+export function nameFromCommand(command: ExternalCommand | ExternalCommandDefinition, name?: string): string {
```
taylorhadden, line 33: In place of some of this you can use `rawOrStoreValue()` in `ReadableStore.ts`.

apps/tangent-electron/src/main/messages/index.ts
```diff
@@ -19,30 +20,76 @@ import './themes'
...
+		})
+		
+		const processTimeout = 5 * 1000
+        let timedOut = false
+        let stdout = ''
+        let stderr = ''
+		
+        const timer = setTimeout(() => {
+            timedOut = true
+            log.info('Process is taking too much')
```
taylorhadden: It would be good to log the length of time that was timed out.

apps/tangent-electron/src/main/messages/index.ts
```diff
@@ -19,30 +20,76 @@ import './themes'
...
+            const result = {
+				stdout: stdout,
+                stderr: stderr,
+                exitCode: code
+            }
+			
+			windowHandle.postUserMessage(
+				code === 0 ? 'info' : 'warning', 
+				code === 0 ? 'Command Completed' : 'Command Failed',
+				code === 0 ? stdout : stderr)
```
taylorhadden: This could pump out a _ton_ of text for some complex commands. Probably not needed for now?

apps/tangent-electron/src/main/messages/index.ts
```diff
@@ -19,30 +20,76 @@ import './themes'
...
+                exitCode: 1
+            }
+			windowHandle.postUserMessage('error', 'Command Failed', stderr)
+			log.error(`Process error: ${err.message}`, result)
+            resolve(result)
+        })
+    })
 })
 
+
```
taylorhadden: Nit: unneeded extra space.

apps/tangent-electron/package.json
```diff
@@ -41,6 +41,7 @@
     "email": "taylorhadden@icloud.com"
   },
   "dependencies": {
+    "@ericcornelissen/lregexp": "^1.0.13",
```
taylorhadden, line 44: I didn't see this used. Am I blind?

apps/tangent-electron/src/app/model/commands/ExecCliCommand.ts
```diff
@@ -0,0 +1,95 @@
...
+
+		// return actions
+	// }
+
+	getName() {
+		return 'Execute CLI'
+	}
+
+	getLabel(){
+		return 'Execute Commands'
```
taylorhadden: This probably should be "Execute External Commands."

apps/tangent-electron/src/app/views/external-commands/ExternalCommandEditor.svelte
```diff
@@ -0,0 +1,162 @@
...
+			<p class={'explanation ' + message.level}>{@html message.message}</p>
+		{/each}
+	{/if}
+	<details>
+		<summary>Command Template Tokens</summary>
+		<br>
+		<p>These tokens are replaced with the appropriate values when command is called.</p>
+		<table><tbody>
+			{#each commandTemplates as commandTemplate}
+				<tr><th><ExternalCommandTemplateButton templateText={commandTemplate.text} insertTemplateText={insertTextIntoTemplate}/></th><td>{commandTemplate.description}</td></tr>
```
taylorhadden, line 90: I see that the old button was removed but I don't see a replacement. Should this be the other template button?

taylorhadden (changes requested):

apps/tangent-electron/package.json
hamidb80: No you're not, this is dependency of `shescape`, See this issue https://github.com/ericcornelissen/shescape/issues/2619

`shescape` is used to escape shell arguments

apps/tangent-electron/src/app/views/external-commands/ExternalCommandEditor.svelte
hamidb80: I've used the button for creation rules.
```ts
import ExternalCommandTemplateButton from '../creation-rules/CreationRuleTemplateButton.svelte'
```

apps/tangent-electron/src/app/model/commands/ExecuteExternalCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+	workspace.api.os.execCLI(command.commandTemplate, context)
+}
+
+interface ExecuteExternalCommandContext extends CommandContext {
+	command?: ExternalCommand
+}
+
+export default class ExecuteExternalCommand extends WorkspaceCommand {
+	constructor(workspace: Workspace) {
+		super(workspace, { group: 'Notes' }) // <---- this makes it available when editing notes
```
taylorhadden, line 27: Having an empty group should be fine, and is probably what is desired. The groups are used to separate out groups of bindings that won't ever be in scope with each other.

apps/tangent-electron/src/app/model/commands/index.ts
```diff
@@ -51,7 +52,7 @@ export type WorkspaceCommands = LiteralCommands & GenericCommands
 
 function createAllCommands(workspace: Workspace) {
 	return {
-
+
```
taylorhadden, line 55: Nit: this is just a whitespace change.

apps/tangent-electron/src/app/views/external-commands/ExternalCommandEditor.svelte
```diff
@@ -0,0 +1,162 @@
...
+<main>
+	<header>
+		<slot name="header-left"></slot>
+		<!-- svelte-ignore a11y-missing-content -->
+		<h2 class="name"
+			use:editable={commandName}
+			use:tooltip={"Define the name of the command. Set an emoji as the first character of the name to make an icon."}
+		></h2>
+	</header>
+	<label use:tooltip={"Defines how the note will be named. Refer to the Template Token list for available dynamic values."}>
```
taylorhadden, line 75: This tooltip and label span are still mentioning the creation rules. It's also bound to the same value as the header. Seems like we don't need both. Is one of these supposed to be the command template?

taylorhadden (changes requested):

hamidb80:
Hi, sorry for the delay

Demo so far
image

apps/tangent-electron/src/app/model/commands/ExecuteExternalCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+		}
+		return actions
+	}
+
+	getName() {
+		return 'Execute Commands'
+	}
+
+	getLabel(){
+		return 'Execute External Commands'
```
taylorhadden: Ideally the label and tooltip would be able to share the name and any description of an external command set within the context.

apps/tangent-electron/src/main/messages/index.ts
```diff
@@ -19,8 +20,77 @@ import './themes'
...
+            const result = {
+				stdout: stdout,
+                stderr: stderr,
+                exitCode: code
+            }
+			
+			const commandType =  code === 0 ? 'info' : 'warning'
+			const commandTitle = code === 0 ? 'Command Completed' : 'Command Failed'
+			const commandMsg  = code === 0 ? 'Successfully' : `With error code ${code}. See logs`
+			// console.log({commandType, commandTitle, commandMsg})
```
taylorhadden: Nit: remove the debug comment.

taylorhadden (changes requested):
If you can, just two more tweaks!

apps/tangent-electron/src/app/model/commands/ExecuteExternalCommand.ts
hamidb80: do you mean like this one? (the tooltip is description of the command)

image

hamidb80: or do you mean to put entire resolved command (like `python3 action.py RESOLVED_FILE_PATH RESOLVED_WORKDIR_PATH` ) ?

taylorhadden (approved):

apps/tangent-electron/src/app/model/commands/ExecuteExternalCommand.ts
taylorhadden: Exactly what you're showing in the picture! (But keep the double quotes you just added).

taylorhadden:
This is looking good, it just needs to be resolved against the versioning churn in `main`. I wouldn't recommend manually merging `package-lock`. I would revert the changes and run `npm install` again.

Issue #370, opened 2026-05-31
hierarchical attachments directories


Feisquared2:
Hi

# Problem
I often write reports at work, most of them unrelated to each other. having global attachment folder makes me little conformable since I have to move the linked attachments manually to a near directory.

# Possible Solutions

1. open a separate workspace from report directory; but then I cannot access my other notes in parent directories...
3. making Tangent itself support this feature
4. having `.tangentconf` that rewrites global configurations in directories
5. use custom script:  with #367 I can write a simple script to move them to target directory. 
6. custom fields in **front matter** like `attachment_dir: ...`
----

I like (4), (5), (2)

taylorhadden:
One of the things I've thought about is having something like a relative attachments location, but that wouldn't support a situation where you want all items in a particular folder to put their attachments local to that folder.

Number 3 actually makes the most sense to me for this. The config file already exists as `.tangentfolder`.

Feisquared2:
> Number 3 actually makes the most sense to me for this. The config file already exists as `.tangentfolder`.

well, wouldn't it increase number of I/O operations (by a lot)? 
I guess the process goes like: 
1. go to parent directory
2. check for `.tangentfolder`
3. read the file
4. repeat

in my case the process would repeat 4 times on average

taylorhadden:
I'm 100% not worried about it at the scale of pasting images into a note.

hamidb80:
Hi Taylor 
do you have any ideas for this? 

maybe in my spare time I could work on this

----

may I ask for multi-workspace-paste directory? 

like in the settings you could write `./figs/, ./**/assets/, last-fallback-dir/`

it seems I only need to modify `getAttachmentPath` function

taylorhadden:
Actually, the attachment location _really should be_ workspace-specific. In my ideal world, there is a way to customize every setting globally or per-workspace, VSCode style.

If you want to take a crack at it, go ahead! If you want to avoid having a per-workspace settings layer, you could focus on adding attachment settings to the `FolderInfo` files and using that to set the attachment path.

PR #389, opened 2026-06-19
 multiple path choices for pasted image

hamidb80:
addresses #370

### DIFF
--- a/apps/tangent-electron/src/common/WindowApi.ts
+++ b/apps/tangent-electron/src/common/WindowApi.ts
@@ -6,6 +6,7 @@ import type { ContextMenuTemplate } from './menus'
 import type { QueryResult } from './indexing/queryResults'
 import type { QueryParseResult } from '@such-n-such/tangent-query-parser'
 import type { UrlData } from './urlData'
+import type { AttachmentRuleDefinition } from './settings/AttachmentRule'
 
 export interface SelectPathOptions {
 	title?: string
@@ -60,7 +61,7 @@ export default interface WindowAPI {
 	system: {
 		getAllFonts(): Promise<string[]>
 		getAllLanguages(): Promise<string[]>
-		saveImageFromClipboard(contextPath: string): Promise<string>
+		saveImageFromClipboard(contextPath: string, attachmentRules: AttachmentRuleDefinition[]): Promise<string>
 		copyImageToClipboard(path: string): Promise<void>
 		updateImageFromClipboard(path: string): Promise<void>
 		messageDialog(options: Electron.MessageBoxOptions): Promise<Electron.MessageBoxReturnValue>
@@ -114,7 +115,7 @@ export default interface WindowAPI {
 		/** Open a url in the default format */
 		openExternal(path: string)
 		getTitle(href: string): Promise<string>
-		saveFromUrl(href: string, contextPath: string): Promise<string>
+		saveFromUrl(href: string, contextPath: string, attachmentRules: AttachmentRuleDefinition[]): Promise<string>
 		getUrlData(url: string): Promise<UrlData>
 	}
 
--- a/apps/tangent-electron/src/common/settings/AttachmentRule.ts
+++ b/apps/tangent-electron/src/common/settings/AttachmentRule.ts
@@ -4,7 +4,7 @@ import { isEmptyOrWhitespace } from 'common/stringUtils'
 
 export type ResolveMode = 'absolute' | 'relative' | 'upward'
 
-export type AttachmentRulerDefinition = AttachmentRule | AttachmentRuleDefinition
+export type AttachmentRulesDefinition = AttachmentRule | AttachmentRuleDefinition
 
 export interface AttachmentRuleDefinition {
 	name: string
--- a/apps/tangent-electron/src/common/settings/AttachmentRule.ts
+++ b/apps/tangent-electron/src/common/settings/AttachmentRule.ts
@@ -2,23 +2,18 @@ import { ObjectStore, type ObjectStoreOptions, WritableStore, ValidatingStore, r
 import Setting, { type SettingDefinition } from './Setting'
 import { isEmptyOrWhitespace } from 'common/stringUtils'
 
-export type ResolveMode = 'absolute' | 'relative' | 'upward'
+export type ResolveMode = 'default' | 'upward'
 
 export type AttachmentRulesDefinition = AttachmentRule | AttachmentRuleDefinition
 
 export interface AttachmentRuleDefinition {
-	name: string
 	resolveMode: ResolveMode
 	path: string
 	description: string
 }
 
 export function nameFromRule(rule: AttachmentRule | AttachmentRuleDefinition, defaultValue: string): string {
-	return (typeof rule.name == 'string' ? rule.name : rule.name.value) || defaultValue
-}
-
-export function willPromptForName(name: string) {
-	return isEmptyOrWhitespace(name)
+	return (typeof rule.path == 'string' ? rule.path : rule.path.value) || defaultValue
 }
 
 const pathDefinition: SettingDefinition<string> = {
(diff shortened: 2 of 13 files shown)


apps/tangent-electron/src/main/Workspace.ts
```diff
@@ -1088,25 +1088,56 @@ export default class Workspace {
...
+						log.info('Relative attachment path would have brought us outside the workspace. Will use the root instead. Original:', targetDirectory)
+						targetDirectory = this.rootPath
+					}
+				}
+				else {
+					targetDirectory = path.join(this.rootPath, targetDirectory)
+				}
 			}
+
+			break
```
taylorhadden: Instead of invariably breaking here and conditionally continuing on line 1122, I think it makes more sense to break if `targetDirectory` is valid here.

apps/tangent-electron/src/main/Workspace.ts
```diff
@@ -1088,25 +1088,56 @@ export default class Workspace {
 	}
 
 	getAttachmentPath(idealFilepath: string, contextPath: string): string {
-		const contextDir = path.dirname(contextPath) // TODO: Use this
-		let targetDirectory = getSettings().defaultPasteLocation.value
-		if (!targetDirectory) {
-			targetDirectory = this.rootPath
+		// tries to find the closest local folder indicated by wildcard `./**/assets/dir`
+		// it also supports fallback
+		// for example:  ./**/figs/, global-assets/
```
taylorhadden: What happens if the folder doesn't exist? Does somebody have to create a "figs" folder first?

The convenience of automatically creating the appropriate folder is really useful.

apps/tangent-electron/src/main/Workspace.ts
```diff
@@ -1088,25 +1088,56 @@ export default class Workspace {
...
-				targetDirectory = path.resolve(path.join(contextDir, targetDirectory))
-				if (!targetDirectory.startsWith(this.rootPath)) {
-					log.info('Relative attachment path would have brought us outside the workspace. Will use the root instead. Original:', targetDirectory)
-					targetDirectory = this.rootPath
-				}
+
+		const contextDir = path.dirname(contextPath) // TODO: Use this
+		const targetDirectories = getSettings().defaultPasteLocation.value.split(',').map(s => s.trim())
+		let	targetDirectory = null
+		const local_wild_card = "./**/"
```
taylorhadden: This variable should be camelCased.

apps/tangent-electron/src/main/Workspace.ts
```diff
@@ -1088,25 +1088,56 @@ export default class Workspace {
...
+		// it also supports fallback
+		// for example:  ./**/figs/, global-assets/
+
+		function findUp(startPath: string, target: string, rootPath: string) {
+			let currentPath = path.resolve(startPath)
+			const root = path.resolve(rootPath)
+			
+			while (true) {
+				const targetPath = path.join(currentPath, target)
+				if (fs.existsSync(targetPath)) return targetPath
```
taylorhadden: Sync file actions lock down the entire thread. It would be better to make the main and inner functions async and update the call sites. Both places where this is used can easily be converted to async functions.

apps/tangent-electron/src/main/Workspace.ts
taylorhadden: Is "local" the right name here? It's looking for the closest sibling/ancestor, right?

taylorhadden (changes requested):
This seems like a pretty useful feature. We should also update the documentation and the description of the wildcard in the Attachments settings.

hamidb80:
Notes
* error handling for path errors, etc should be added in the UI
* someone may want to have comma `,` in the directory's name for that we should allow `"dir,name"` inside `"` but maybe that's overthinking...

apps/tangent-electron/src/main/Workspace.ts
```diff
@@ -1087,26 +1087,42 @@ export default class Workspace {
...
-				targetDirectory = path.resolve(path.join(contextDir, targetDirectory))
-				if (!targetDirectory.startsWith(this.rootPath)) {
-					log.info('Relative attachment path would have brought us outside the workspace. Will use the root instead. Original:', targetDirectory)
-					targetDirectory = this.rootPath
-				}
+		const targetDirectories = [
+			...getSettings().defaultPasteLocation.value.split(',').map(s => s.trim()), // paths are separated by comma (,)
+			'' // default fallback to root path
+		] 
+		const closestWildCardPrefix = "./**/"
```
taylorhadden: A key problem with this approach is that this actually normally has an inverted meaning. This should technically be "any file that is a sibling or _lower_" by standard globstar syntax.

There are many options that we might want for finding & creating attachment folders, and they are not something easily encoded into a string. I think it would be better to have each option be an object, similar to how the Creation Rules are defined. That way we can define things like "find nearest ancestor" or "create if not found" in each option.

apps/tangent-electron/src/main/files.ts
```diff
@@ -53,3 +53,53 @@ export async function loadTreeFromPath(
...
+ * Searches upward through the file system hierarchy for a target file or directory.
+ * 
+ * @param rootAbsPath - The absolute path of root directory
+ * @param startAbsPath - The absolute path of directory starting 
+ * @param targetRelPath - The relative path of the file or directory to search for 
+ * 
+ * @returns The absolute path to the target if found or `null`
+ */
+export async function findUp(rootAbsPath: string, startAbsPath: string, targetRelPath: string): Promise<string | null> {
+	const candidateDirs = findUpCandidates(rootAbsPath, startAbsPath, targetRelPath)
```
taylorhadden, line 90: My other comment may make this approach no longer needed, but we _could_ make `findUpCandidates` a generator function to only create each path as-needed.

taylorhadden (changes requested):
After looking at this some more, I'm no longer thinking that a flat array of strings is the way to tackle this.

apps/tangent-electron/src/main/Workspace.ts
hamidb80: agreed, I'll work on it

hamidb80: honestly I had a weird feeling about this pattern too

hamidb80: honestly I had a weird feeling about this pattern too


image

something like that?

taylorhadden: This is definitely the right direction. I think following a pattern similar to Creation Rules (a list that opens an individual editor) would be even better.

hamidb80:
Taylor do you think how many modes do we need? ( in settings )

1. Default (detects based on ./ or /), upward
2. Absolute, local, upward

hamidb80:
why are there 2 places for settings? 

```js
	settings: Settings
	workspaceSettings = new WritableStore<WorkspaceSettings>(null)
```

the first one is available in `getSettings()` but it's a little tricky to get `workspaceSettings`

apps/tangent-electron/src/main/Workspace.ts
hamidb80: it works!!!

image

image

hamidb80:
it now works but the UI it's a little messy

Taylor please checkout and give me feedback. do you agree with the direction?

apps/tangent-electron/src/common/WindowApi.ts
```diff
@@ -60,7 +61,7 @@ export default interface WindowAPI {
 	system: {
 		getAllFonts(): Promise<string[]>
 		getAllLanguages(): Promise<string[]>
-		saveImageFromClipboard(contextPath: string): Promise<string>
+		saveImageFromClipboard(contextPath: string, attachmentRules: AttachmentRuleDefinition[]): Promise<string>
```
taylorhadden, line 64: You shouldn't need to route the rules through here. The workspace settings should theoretically be accessible from the main process.

As I check, I can see I have a `TODO` about this in the `main/Workspace.ts` file. It uses the same `DataFile` concept as the renderer process, so it should be relatively straightforward.

apps/tangent-electron/src/common/settings/AttachmentRule.ts
```diff
@@ -0,0 +1,82 @@
...
+const pathDefinition: SettingDefinition<string> = {
+	name: 'path',
+	description: 'the folder want your attachments paste in', 
+	defaultValue: '',
+	form: 'folder'
+}
+
+const attachmentModeDefinition: SettingDefinition<ResolveMode> = {
+	name: 'Creation Mode',
+	description: 'Determines how the new note is created.',
```
taylorhadden, line 28: The descriptions here haven't been updated.

apps/tangent-electron/src/common/settings/AttachmentRule.ts
```diff
@@ -0,0 +1,82 @@
...
+
+// Used for differentiation in drag/drop
+// TODO: Maybe these could be persistent? Maybe they can help with list sync?
+let AttachmentRuleID = 0
+
+const creationStoreOptions: ObjectStoreOptions = {
+	patchBlockList: ['id']
+}
+
+export default class AttachmentRule extends ObjectStore {
```
taylorhadden, line 60: Should the attachment rule also define how the pasted attachment is named? Is the current naming scheme sufficient?

Having to redefine the naming template might be a pain though.

taylorhadden (changes requested):
This looks mostly like the right direction. You shouldn't need to hand over the attachment rules across the renderer/main divide though. We should make those settings accessible.

That could be something that I do separately if you prefer.

PR #392, opened 2026-06-23
fix RTL line gutter issue

hamidb80:
fix a problem where the `copy` button intersects with the RTL TODO checkbox

## before
image

## after 
image

---

```md
- [] hello
	- [ ] item 1
	- [ ] item 2
- [ ] سلام
	- [ ] مورد 1
	- [ ] مورد 2
```

### DIFF
--- a/.github/actions/setup-and-build/action.yml
+++ b/.github/actions/setup-and-build/action.yml
@@ -1,0 +1,18 @@
+name: Setup and Build
+description: Checks out the repo, sets up Node, installs dependencies, and builds.
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/setup-node@v5
+      with:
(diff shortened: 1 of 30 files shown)


hamidb80:
Oh no issue when there is no check box

## without check box
image

## with checkbox
image

taylorhadden:
Two comments:
- The commit history of this branch is a little borked. I would recommend starting a new branch and cherry picking the relevant commits. Generally, best practice is to always start from a clean branch. Prefer rebasing a pull request to merging in main.
- Have you tried [inset-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-start) instead of `left`? That may be the best solution.

Issue #325, opened 2026-02-19
Ability to zoom in PDFs like in images


hamidb80:
Hi,

a drawer shows up when I scroll the mouse wheel in images

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/8f40ffa6-ce67-4a2e-b1c7-2f2dc372da37" />

but in PDFs only this shows up.

<img width="1582" height="1116" alt="Image" src="https://github.com/user-attachments/assets/d65e08b3-6b3b-4fae-822b-57dc16472cf5" />

Feisquared2:
Hi, can this be implemented?

foodgellow:
I thought the other day that Tangent in note-taking world is like Zotero in PDF world. 

Taylor have you used it? it's amazing the way it manages comments and references. I'd wish the 2 combined

taylorhadden:
I have not. I'm not in the research paper world, so it's never been something hitting my niche.

hamidb80:
> I have not. I'm not in the research paper world, so it's never been something hitting my niche.

in Zotero you can also download a webpage and highlight it. and then just by you can search among all of your highlights and comments. pretty neat IMAO, but kinda differs from purpose of PKM

taylorhadden:
Implemented in v0.12.2.

Thank you @hamidb80!

PR #393, opened 2026-06-24
PDF zoom functionality

hamidb80:
addresses #325 

> ~~thanks to DeepSeek for the relative position functionality~~

### DIFF
--- a/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
+++ b/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+import { fly } from 'svelte/transition'
 import { getContext, onDestroy } from 'svelte'
 import { Workspace } from 'app/model'
 import PdfViewState from 'app/model/nodeViewStates/PdfViewState'
@@ -30,26 +31,57 @@ let viewerElement: HTMLDivElement
 let viewer: pdfviewer.PDFViewer = null
 
 let zoom = state.zoom
-let panX = state.panX
-let panY = state.panY
+
+let isHoveringControls = false
+$: isTransformed = $zoom !== 1
 
 function onWheel(event: WheelEvent) {
 	container.focus()
 	if (event.ctrlKey) {
-		event.preventDefault()
-		zoom.applyWheelEvent(event)
-		
-		viewer.currentScale = $zoom
+		event.preventDefault();
+
+		let somePageElement = container.querySelector('.page');
+
+		// Get mouse position relative to element BEFORE zoom
+		const rectBefore = somePageElement.getBoundingClientRect();
+		const mouseX = event.clientX;
+		const mouseY = event.clientY;
+
+		// Calculate mouse position in page coordinates (before zoom)
+		const pageXBefore = (mouseX - rectBefore.left) / rectBefore.width;
+		const pageYBefore = (mouseY - rectBefore.top) / rectBefore.height;
+
+		// Apply zoom
+		zoom.applyWheelEvent(event);
+		viewer.currentScale = $zoom;
+
+		// Get element AFTER zoom
+		const rectAfter = somePageElement.getBoundingClientRect();
+
+		// Calculate where the mouse should be in page coordinates (same position)
+		const pageXAfter = pageXBefore;  // We want the same relative position!
+		const pageYAfter = pageYBefore;
+
+		// Calculate the target scroll position to keep mouse fixed
+		const targetScrollX = (pageXAfter * rectAfter.width) + rectAfter.left - mouseX;
+		const targetScrollY = (pageYAfter * rectAfter.height) + rectAfter.top - mouseY;
+
+		// Apply scroll to keep mouse position fixed
+		container.scrollLeft += targetScrollX;
+		container.scrollTop += targetScrollY;
 	}
 	else if ($zoom !== 1) { // Helps when in tangent view
 		event.preventDefault()
-		$panX = $panX + event.deltaX * +1 * (1/$zoom)
-		$panY = $panY + event.deltaY * +1 * (1/$zoom)
-
-		container.scrollLeft = $panX; // horizontal pan
-  		container.scrollTop = $panY; // vertical pan
+		container.scrollLeft += event.deltaX * +1 * (1/$zoom) // horizontal pan
+  		container.scrollTop += event.deltaY * +1 * (1/$zoom) // vertical pan
 	}
 }
+
+function resetTransform() {
+	$zoom = 1
+	viewer.currentScale = 1
+}
+
 
 async function doPDF() {
 	let pdf = await pdfjs.getDocument(state.file.cacheBustPath).promise
@@ -156,6 +188,15 @@ function onClick(event: MouseEvent) {
 			<div bind:this={viewerElement} on:click={onClick}></div>
 		</div>
 	</article>
+
+	{#if isTransformed || isHoveringControls}
+		<div class="controls" transition:fly={{ y: 100 }}
+			style:bottom={`calc(1em + ${extraBottom}px)`}
+		>
+			<button class="zoomText" on:click={resetTransform}>{Math.round($zoom * 100)}%</button>
+			<input class="zoomSlider" type="range" min="{zoom.range.min}" max={zoom.range.max} step="0.1" bind:value={$zoom}/>
+		</div>
+	{/if}
 </main>
 
 <style lang="scss">
@@ -188,6 +229,18 @@ article {
 		-webkit-user-select: text;
 		user-select: text;
 	}
+}
+
+
+.controls {
+	position: absolute;
+	z-index: 1;
+	left: 1em;
+	bottom: 1em;
+
+	display: flex;
+	flex-direction: row;
+	gap: .5em;
 }
 
 </style>
--- a/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
+++ b/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
@@ -9,7 +9,6 @@ import * as pdfviewer from 'pdfjs-dist/web/pdf_viewer.mjs'
 import { resizeObserver } from 'app/utils/resizeObserver'
 import { scrollTo } from 'app/utils'
 import { smoothScrollTime } from 'app/utils/style'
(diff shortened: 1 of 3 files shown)


hamidb80:
Taylor I've copied these lines from `ImageView.svelte`, is that right or I should move them to a global style file?

```css

.controls {
	position: absolute;
	z-index: 1;
	left: 1em;
	bottom: 1em;

	display: flex;
	flex-direction: row;
	gap: .5em;
}
```

hamidb80:
I wanna hide navigation in `focus mode` in PDF viewer (like in notes), what should I do?

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -29,6 +30,59 @@ let viewerElement: HTMLDivElement
...
+		const targetScrollX = (pageXAfter * rectAfter.width) + rectAfter.left - mouseX;
+		const targetScrollY = (pageYAfter * rectAfter.height) + rectAfter.top - mouseY;
+
+		// Apply scroll to keep mouse position fixed
+		container.scrollLeft += targetScrollX;
+		container.scrollTop += targetScrollY;
+	}
+	else if ($zoom !== 1) { // Helps when in tangent view
+		event.preventDefault()
+		container.scrollLeft += event.deltaX * +1 * (1/$zoom) // horizontal pan
```
taylorhadden: Nitpick: no need for the extra `+1` term in these.

Did you find that scrolling was too fast if zoom wasn't taken into account? I would have expected the system to handle that.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -29,6 +30,59 @@ let viewerElement: HTMLDivElement
...
+
+		// Calculate the target scroll position to keep mouse fixed
+		const targetScrollX = (pageXAfter * rectAfter.width) + rectAfter.left - mouseX;
+		const targetScrollY = (pageYAfter * rectAfter.height) + rectAfter.top - mouseY;
+
+		// Apply scroll to keep mouse position fixed
+		container.scrollLeft += targetScrollX;
+		container.scrollTop += targetScrollY;
+	}
+	else if ($zoom !== 1) { // Helps when in tangent view
```
taylorhadden: I'm not sure this is exactly what is wanted here. Doesn't this prevent any scrolling action in the thread view while the PDF is scaled?

taylorhadden (changes requested):

taylorhadden:
The mechanism for hiding the top chrome is a bit of a mess. It's invoked on line 638 of `NoteEditor.svelte` and handled on line 52 of `WorkspaceView.svelte`. There's also an interaction with `workspace.viewState.focusing`.

It's a bit of a spider web. I would recommend making it a separate PR.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
hamidb80: yeah? I mean we should capture every Wheel event, it works with mouse, I've not tested on my laptop with "touchpad".

is that what you're asking (to test on touchpad)? or I'm misunderstood

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
hamidb80: on first try zoom seemed fast becuase the starting zoom was not set correctly (it was set to 1, but scale of `PDFView` _actually set according to the page width_)

I guess now it has been resolved

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -29,6 +30,65 @@ let viewerElement: HTMLDivElement
...
+function resetZoom(val: number | 'auto') {
+	viewer.currentScaleValue = `${val}`
+	$zoom = viewer.currentScale
+}
+
+function setZoom(){
+	resetZoom(zoom.value)
+}
+
+function resetZoomEvent(val: Event) {
```
taylorhadden: Convention for event handlers has generally been like "onResetZoom".

This looks like it wouldn't actually reset the zoom to 100%. Is the issue that we don't know what the target "non-scaled" zoom should be?

Perhaps we should handle zoom at a higher level. What if we scaled the size of the `<article>` element instead? Then the PDF zoom handler would just need to fill that space.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
taylorhadden: So does that mean this branch doesn't need to exist at all?

taylorhadden (changes requested):

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
hamidb80: > What if we scaled the size of the <article> element instead? Then 

the PDF zoom handler would just need to fill that space.
the pdfviewer API has `currentScaleValue = 'auto'`

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
hamidb80: yes, no need for this

hamidb80:
I'm investigating new API which is faster

https://github.com/mozilla/pdf.js/issues/18076

The current implementation is so laggy with touchpad

hamidb80:
the control is now very smooth

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -28,6 +28,53 @@ let container: HTMLDivElement
...
+		const [z2, z1] = zoom.applyWheelEvent(event)
+		const containerBB  = container.getBoundingClientRect()
+		const relativeCursorPos = [event.clientX - containerBB.left, event.clientY - containerBB.top]
+	
+		viewer.updateScale({ drawingDelay, scaleFactor: z2/z1, origin: relativeCursorPos })
+		$zoom = parseFloat(viewer._currentScaleValue)
+	}
+	else {
+		event.preventDefault()
+		container.scrollLeft += event.deltaX * +1 * (1 / $zoom)
```
taylorhadden: NIt: the `+1` terms don't do anything.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -28,6 +28,53 @@ let container: HTMLDivElement
...
+	container.focus()
+
+	if (event.ctrlKey) {
+		event.preventDefault()
+
+		const [z2, z1] = zoom.applyWheelEvent(event)
+		const containerBB  = container.getBoundingClientRect()
+		const relativeCursorPos = [event.clientX - containerBB.left, event.clientY - containerBB.top]
+	
+		viewer.updateScale({ drawingDelay, scaleFactor: z2/z1, origin: relativeCursorPos })
```
taylorhadden: What effect does the drawing delay have? Is it so that trackpad scroll events are debounced?

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -28,6 +28,53 @@ let container: HTMLDivElement
...
+	else {
+		event.preventDefault()
+		container.scrollLeft += event.deltaX * +1 * (1 / $zoom)
+		container.scrollTop += event.deltaY * +1 * (1 / $zoom)
+	}
+}
+
+function setZoom(val: number | 'auto') {
+	if (val == 'auto') {
+		viewer.currentScaleValue = `${val}`
```
taylorhadden: You should be able to just assign to `val` here.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -28,6 +28,53 @@ let container: HTMLDivElement
...
+		$zoom = parseFloat(viewer._currentScaleValue)
+	}
+	else {
+		event.preventDefault()
+		container.scrollLeft += event.deltaX * +1 * (1 / $zoom)
+		container.scrollTop += event.deltaY * +1 * (1 / $zoom)
+	}
+}
+
+function setZoom(val: number | 'auto') {
```
taylorhadden: Nitpick: I prefer the full `value` or `zoomValue` here.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -28,6 +28,53 @@ let container: HTMLDivElement
...
+		viewer.updateScale({ 
+			drawingDelay, 
+			scaleFactor: val / parseFloat(viewer.currentScaleValue), 
+			origin:  [countainerBB.width / 2 , countainerBB.height / 2]
+		})
+	}
+	$zoom = viewer.currentScale
+}
+
+function onZoomSet(evt: Event) {
```
taylorhadden: Please don't abbreviate whenever possible. In this case, the event is not used and can be omitted entirely.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -28,6 +28,53 @@ let container: HTMLDivElement
...
+
+const drawingDelay = 150
+
+function onWheel(event: WheelEvent) {
+	container.focus()
+
+	if (event.ctrlKey) {
+		event.preventDefault()
+
+		const [z2, z1] = zoom.applyWheelEvent(event)
```
taylorhadden: It would be clearer for these to be named `newZoom` and `oldZoom`, or something equally clear.

taylorhadden (changes requested):

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
hamidb80: It is the PDFjs API to prevent re-drawing too frequently, think of it as throttle

hamidb80: Like in "Google Maps" when you zoom in but the re-drawing occurs after a small delay

taylorhadden (approved):

taylorhadden:
Awesome, thank you!

PR #394, opened 2026-06-24
fix `copy` button for RTL TODO list ( right `LineGutter` )

hamidb80:
fix a problem where the `copy` button intersects with the RTL TODO checkbox

## before
image

## after 
image

```md
- [] hello
	- [ ] item 1
	- [ ] item 2
- [ ] سلام
	- [ ] مورد 1
	- [ ] مورد 2
```

--------

https://github.com/suchnsuch/Tangent/pull/392#issuecomment-4785123252

### DIFF
--- a/apps/tangent-electron/src/app/views/editors/NoteEditor/LineGutter.svelte
+++ b/apps/tangent-electron/src/app/views/editors/NoteEditor/LineGutter.svelte
@@ -30,14 +30,21 @@ function positionOnLine(line: HTMLElement, container: HTMLElement) {
 		placement: side === 'left' ? 'left-start' : 'right-start'
 	}).then(result => {
 		const lineStyle = getComputedStyle(line)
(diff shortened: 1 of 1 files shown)


hamidb80:
> @taylorhadden Have you tried [inset-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset-inline-start) instead of left? That may be the best solution.

Ummm.. the problem is that line gutter is positioned manually and it's not something itself contains RTL or LTR characters. 

did I understand the property correctly? I think it's not relevant at least here.

I've also tried, but didn't work
```js
const marginLeft = parseFloat(lineStyle.marginLeft)

container.style.direction = side === 'left' ? 'ltr' : 'rtl'
container.style.insetInlineStart = result.x - marginLeft + 'px'
```

taylorhadden (approved):
Now I understand what you're doing! Thanks for the fix!

PR #408, opened 2026-07-29
PDF: add pan across x axis by pressing space and moving the mouse

kenekoba:
Hi,

almost all PDF viewer softwares allow the user to scroll x axis by pressing space and moving (dragging) the mouse

this feature is not needed in laptops because of touchpad but in desktop that only has mouse & keyboard it's needed

### DIFF
--- a/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
+++ b/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
@@ -207,7 +207,7 @@ function onClick(event: MouseEvent) {
 	<article use:resizeObserver={onResize}>
 		<div class="container pdfViewer" bind:this={container}>
 			<!-- svelte-ignore a11y-click-events-have-key-events -->
-			<!-- svelte-ignore a11y-no-static-element-interactions --> 
+			<!-- svelte-ignore a11y-no-static-element-interactions -->
 			<div bind:this={viewerElement} class={{ 'panning': isPanning }} on:mousemove={onMouseMove} on:click={onClick}></div>
 		</div>
 	</article>
--- a/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
+++ b/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
@@ -48,8 +48,12 @@ function onWheel(event: WheelEvent) {
 	}
 	else {
 		event.preventDefault()
-		container.scrollLeft += event.deltaX * (1 / $zoom)
-		container.scrollTop += event.deltaY * (1 / $zoom)
+		const dx = event.deltaX * (1 / $zoom)
+		const dy = event.deltaY * (1 / $zoom)
+
+		// shift key change direction of scroll
+		container.scrollLeft += event.shiftKey ? dy : dx 
+		container.scrollTop += event.shiftKey ? dx : dy
 	}
 }
 
@@ -64,7 +68,7 @@ function onMouseMove(event: MouseEvent) {
 }
 
 function onKeyDown(event: KeyboardEvent) {
-	if (event.code == "Space"){
+	if (event.key == "Space"){
 		event.preventDefault()
 		isSpacePressed = true
 		container.style.cursor = 'grab'
--- a/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
+++ b/apps/tangent-electron/src/app/views/node-views/PdfView.svelte
@@ -53,6 +53,28 @@ function onWheel(event: WheelEvent) {
 	}
 }
 
+$: isSpacePressed = false
+
+function onMouseMove(event: MouseEvent) {
+	if (isSpacePressed){
+		event.preventDefault()
+		container.scrollLeft -= event.movementX * (1 / $zoom)
+		container.scrollTop -= event.movementY * (1 / $zoom)
+	}
+}
+
+function onKeyDown(event: KeyboardEvent) {
+	if (event.code == "Space"){
+		event.preventDefault()
+		isSpacePressed = true
+		container.style.cursor = 'grab'
+	}
+}
+function onKeyUp(event: KeyboardEvent) {
+	isSpacePressed = false
+	container.style.cursor = 'auto'
+}
+
 function setZoom(zoomValue: number | 'auto') {
 	if (zoomValue == 'auto') {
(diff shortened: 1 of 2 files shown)


apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -53,6 +53,28 @@ function onWheel(event: WheelEvent) {
...
+function onMouseMove(event: MouseEvent) {
+	if (isSpacePressed){
+		event.preventDefault()
+		container.scrollLeft -= event.movementX * (1 / $zoom)
+		container.scrollTop -= event.movementY * (1 / $zoom)
+	}
+}
+
+function onKeyDown(event: KeyboardEvent) {
+	if (event.code == "Space"){
```
taylorhadden: This should use `event.key` instead.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -53,6 +53,28 @@ function onWheel(event: WheelEvent) {
...
+		event.preventDefault()
+		isSpacePressed = true
+		container.style.cursor = 'grab'
```
taylorhadden: Styling should be driven by Svelte with [class directives](https://svelte.dev/docs/svelte/class) and some css that sets cursor according to that class.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -53,6 +53,28 @@ function onWheel(event: WheelEvent) {
...
+	}
+}
+function onKeyUp(event: KeyboardEvent) {
+	isSpacePressed = false
```
taylorhadden: This should be gated by checking that space is released.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -53,6 +53,28 @@ function onWheel(event: WheelEvent) {
...
+	container.style.cursor = 'auto'
```
taylorhadden: Again, class directives should be used instead.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -53,6 +53,28 @@ function onWheel(event: WheelEvent) {
 	}
 }
 
+$: isSpacePressed = false
```
taylorhadden: State should be about what action it will drive rather than what buttons triggered it. In this case, I would recommend "isPanning" or something similar.

taylorhadden (changes requested):
My primary thought here is: should we instead be panning with a Right click & drag or with a Space + LMB & Drag?

kenekoba:
> My primary thought here is: should we instead be panning with a Right click & drag or with a Space + LMB & Drag?

we may need to reserve right click operations for future features (e.g. highlight, other operations, etc)

kenekoba:
Oh the file name `A note with an incredibly long name...md` is too long and Windows nags about it. I cannot `git revert` and restore the file ...

{7B88001D-8C7B-494D-85DD-14DA34F5B668}

I don't know what to do

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
```diff
@@ -219,6 +249,10 @@ article {
 
 		-webkit-user-select: text;
 		user-select: text;
+
+		.panning {
```
taylorhadden, line 253: You need to use `&.panning` here to have it affect the container as well. This selector compiles to `article .container .panning` and using `&` gives you `article .container.panning`.

test-files/TestWorkspace/Integration Tests/A note with an incredibly long name that's much too long—The length is absolutely insane it's going to go on forever and ever and ever and ever and ever and ever.md
```diff
@@ -1 +1 @@
-And it's got nothing in it!
```
taylorhadden, line 1: Not sure why this is showing up as changed. Ideally it's not in the PR.

taylorhadden: Oh, I see your other comment now. That's annoying! Windows' terrible file management issues rise again. I should shorten that filename…

taylorhadden (changes requested):
Thanks for making these changes! Just a small note about the css.

I'll see how Space + Drag feels. I anticipate that I will change it to be a Space + LMB + Drag gesture, but we'll see.

apps/tangent-electron/src/app/views/node-views/PdfView.svelte
kenekoba: the `panning` class is applied to the `viewerElement` not the `container` itself:

```svelte
		<div class="container pdfViewer" bind:this={container}>
			<!-- svelte-ignore a11y-click-events-have-key-events -->
			<!-- svelte-ignore a11y-no-static-element-interactions -->
			<div bind:this={viewerElement} class={{ 'panning': isPanning }} on:mousemove={onMouseMove} on:click={onClick}></div>
		</div>

```

test-files/TestWorkspace/Integration Tests/A note with an incredibly long name that's much too long—The length is absolutely insane it's going to go on forever and ever and ever and ever and ever and ever.md
kenekoba: the stupid part is that I've managed to restore the file by Github web editor, but it's still listed as changed files, despite that I've copied the file content exactly as it was

{BA8A0F9C-4338-4BA2-B6F9-B7B2691E7EFF}

kenekoba:
> I'll see how Space + Drag feels. I anticipate that I will change it to be a Space + LMB + Drag gesture, but we'll see.

OK, I think I'm done here. 
choose what feels more natural

as a reminder, I've added another feature where pressing shift while scrolling changes the scroll direction. (web browsers implement this feature)

PR #413, opened 2026-08-08
add help for first build :: `git submodule init`

kenekoba:
Hi 

I know that's shame but for noobies like me it may take some time to figure out you need to `init` before `update`


taylorhadden:
Ah, good call. My tooling does all of this automatically. I'll add this myself to tweak the wording.

PR #414, opened 2026-08-08
add rotate scroll when shift key is pressed for image view

kenekoba:
Hi again Taylor!

following my changes for PDF (https://github.com/suchnsuch/Tangent/pull/408) I've added `shift` key scroll behavior for image view too. 
this feature is implemented in most softwares and it is specially useful on desktop PC where touch pad is not available.


taylorhadden (approved):

taylorhadden:
Thanks!

PR #418, opened 2026-08-08
Fix wiki header links with padded labels

### LINKED ISSUE #417: extra space before label and after header, makes header link not working
Hi

when you click on `[[note#header]]` it jumps to `header` in the `note` but this would not work if you write `[[note#header | label]]` but works with `[[note#header| label]]` (note the space between `|`)

alloutflo:
Closes #417.

## Summary

- Trim surrounding whitespace from content IDs inside `createContentIdMatcher()` before the existing block/header matching branches.
- Keep `NoteViewState` passing the parsed `link.content_id` directly to the matcher; use `safeHeaderLine()` only for actual Markdown header text.
- Preserve strict, whole-header `^...$` matching without broadening the regex to surrounding whitespace.
- Add regression coverage for `[[note#My Header | label]]`, header-line trimming, and exact matcher boundaries.
- Ignore macOS `.DS_Store` metadata.

## Root cause

`matchWikiLink` intentionally preserves the source text and returns `My Header ` for the content ID in the reported example. Splitting that raw value produced an empty trailing segment, which made the matcher require a separator after the actual header text.

`createContentIdMatcher()` now trims the lookup ID before building its existing strict matcher. `NoteViewState` passes `link.content_id` directly, while the actual Markdown heading is converted with `safeHeaderLine()`.

## Validation

- Focused header/link tests — 43 passed
- `npx vitest run` — 522 passed, 13 skipped
- `npm run build` — passed
- `git diff --check` — passed

### DIFF
--- a/apps/tangent-electron/src/app/model/nodeViewStates/NoteViewState.ts
+++ b/apps/tangent-electron/src/app/model/nodeViewStates/NoteViewState.ts
@@ -270,7 +270,10 @@ export default class NoteViewState implements NodeViewState, LensViewState {
 			annotation.end = link.end
 		}
 		else if (link.content_id) {
-			const idMatch = createContentIdMatcher(link.content_id)
+			const contentId = link.content_id[0] === '^'
+				? link.content_id
+				: safeHeaderLine(link.content_id)
+			const idMatch = createContentIdMatcher(contentId)
 			if (idMatch) {
 				// Look for headers
 				let start = 0
--- a/apps/tangent-electron/src/app/model/nodeViewStates/NoteViewState.ts
+++ b/apps/tangent-electron/src/app/model/nodeViewStates/NoteViewState.ts
@@ -270,10 +270,7 @@ export default class NoteViewState implements NodeViewState, LensViewState {
 			annotation.end = link.end
 		}
 		else if (link.content_id) {
-			const contentId = link.content_id[0] === '^'
-				? link.content_id
-				: safeHeaderLine(link.content_id)
-			const idMatch = createContentIdMatcher(contentId)
+			const idMatch = createContentIdMatcher(link.content_id)
 			if (idMatch) {
 				// Look for headers
 				let start = 0
--- a/apps/tangent-electron/src/common/markdownModel/links.test.ts
+++ b/apps/tangent-electron/src/common/markdownModel/links.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest'
 
 import { StructureType } from 'common/indexing/indexTypes'
 import { createContentIdMatcher, matchMarkdownLink, matchWikiLink } from './links'
+import { safeHeaderLine } from './header'
 
 describe('Wiki Links', () => {
 	it('Should work for the basics', () => {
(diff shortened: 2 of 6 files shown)


apps/tangent-electron/src/common/markdownModel/links.ts
```diff
@@ -344,16 +344,17 @@ export function resolveLink(store: DefaultIndexStore, link: HrefFormedLink): Tre
...
-		return new RegExp('\^' + contentId.substring(1), 'i')
+		return new RegExp('\^' + normalizedContentId.substring(1), 'i')
 	}
 
 	// Header matches
 	// We want "space-likes" to all be treated the same for cross-compatability & consistency
-	const segments = contentId.split(/[-_ ]+|%20/)
-	return new RegExp('^' + segments.join('([-_ ]|%20)') + '$', 'i')
+	const segments = normalizedContentId.split(/[-_ ]+|%20/)
+	return new RegExp('^ *' + segments.join('([-_ ]|%20)') + ' *$', 'i')
```
taylorhadden: The additional `*` characters is an very aggressive expansion of what a header id should be. It would mean that a small id could match into a longer one that happens to contain the same string. It should not be included.

apps/tangent-electron/src/common/markdownModel/links.test.ts
```diff
@@ -370,6 +370,14 @@ describe('Content ID Matching', () => {
 		expect('My Long Header'.match(createContentIdMatcher('my-long_header'))).toBeTruthy()
 	})
 
+	it('Ignores padding around wiki link header ids and headings', () => {
+		const link = matchWikiLink('[[note#My Header | label]]')
+		const matcher = createContentIdMatcher(link.content_id)
+
+		expect('My Header'.match(matcher)).toBeTruthy()
+		expect(' My Header '.match(matcher)).toBeTruthy()
```
taylorhadden: The matcher should not need to care about surrounding whitespace. The header ID comparison system should instead provide trimmed & converted ids into the matcher.

taylorhadden (changes requested):

apps/tangent-electron/src/common/markdownModel/links.test.ts
taylorhadden: The function for converting arbitrary header text into a content id is `safeHeaderLine()` in `header.ts`.

apps/tangent-electron/src/common/markdownModel/links.test.ts
```diff
@@ -370,8 +371,16 @@ describe('Content ID Matching', () => {
 		expect('My Long Header'.match(createContentIdMatcher('my-long_header'))).toBeTruthy()
 	})
 
+	it('Matches padded wiki link header ids after header normalization', () => {
+		const link = matchWikiLink('[[note#My Header | label]]')
+		const matcher = createContentIdMatcher(safeHeaderLine(link.content_id))
```
taylorhadden: It is incorrect to pass `content_id` through `safeHeaderLine()`. `safeHeaderLine()` is for converting a markdown header line into a valid `content_id`. The trimming behavior should be in `createContentIdMatcher` as well as in `safeHeaderLine()`.

apps/tangent-electron/src/app/model/nodeViewStates/NoteViewState.ts
```diff
@@ -270,7 +270,10 @@ export default class NoteViewState implements NodeViewState, LensViewState {
 			annotation.end = link.end
 		}
 		else if (link.content_id) {
-			const idMatch = createContentIdMatcher(link.content_id)
```
taylorhadden, line 273: I really don't understand what this is trying to do. There should be no other processing needed besides calling `createContentIdMatcher(link.content_id)`.

taylorhadden (changes requested):

apps/tangent-electron/src/common/markdownModel/links.test.ts
alloutflo: You're right. I removed `safeHeaderLine()` from the `content_id` path in 0b4557b. `createContentIdMatcher()` now trims the raw parsed ID, while `safeHeaderLine()` only normalizes the actual Markdown header line. The regression test now passes `link.content_id` directly to the matcher, and the strict `^...$` boundaries remain unchanged. The full suite passes (522 passed, 13 skipped), and the build passes.

apps/tangent-electron/src/app/model/nodeViewStates/NoteViewState.ts
alloutflo: Agreed. In 0b4557b, `NoteViewState` is back to calling `createContentIdMatcher(link.content_id)` directly. The caller-side header/block special case has been removed; trimming is handled inside the matcher. The full suite passes (522 passed, 13 skipped), and the build passes.

taylorhadden (approved):

PR #430, opened 2026-08-22
add `toggle checkbox` command

kenekoba:
Hi,

This PR adds a TODO toggle command that switches between `[ ]` and `[x]`, and vice versa.

The only issue is that when executed on an empty checkbox (`[]`), Tangent panics. Clicking `Retry normal editing` recovers it. All other cases work fine.

Any ideas on what might be causing this?

Thanks!

{CD77A953-9836-428F-9BC5-C0374644EDD8}


kenekoba:
plus, do you think the command should cycle around the states rather than only `on/off`? like 

1. empty
2. checked
3. canceled

apps/tangent-electron/src/app/model/commands/ToDoToggleCommand.ts.ts
```diff
@@ -0,0 +1,116 @@
...
+
+interface ToDoToggleCommandContext extends CommandContext {
+	view?: NoteViewState
+	editor?: MarkdownEditor
+	selection?: EditorRange
+
+}
+
+
+export function toggleToDo(editor: Editor, selection: EditorRange,  mark: string) {
```
taylorhadden: For consistency, it's best for this function to live in `editorActions.ts`

apps/tangent-electron/src/app/model/commands/ToDoToggleCommand.ts.ts
```diff
@@ -0,0 +1,116 @@
...
+	const change = editor.change
+
+	for (const lineRange of doc.getLineRanges(target)) {
+		const [lineStart, lineEnd] = lineRange
+
+		if (lineStart <= cursor && cursor <= lineEnd){
+			const line = doc.getText(lineRange)
+			const match = line.match(listMatcher)
+
+			if (match) {
```
taylorhadden: It would be pretty cool if this could toggle any content into being a checkbox. It should probably add an unchecked checkbox if there isn't one and then just toggle between checked and non-checked from there. The "canceled" state is a pretty special case.

apps/tangent-electron/src/app/model/commands/ToDoToggleCommand.ts.ts
```diff
@@ -0,0 +1,116 @@
...
+				// Index of the first group (the checkbox state)
+				const checkBoxStr = match[8]
+				const checkBoxStart = match.index + match[0].indexOf(checkBoxStr)
+				const head = lineStart + checkBoxStart
+				const tail = head + checkBoxStr.length
+				const cursor = tail - 2
```
taylorhadden: This is redefining `cursor` here. Technically fine, but less confusing to use a different name.

However, it also might want to be avoided entirely. As written, the content of the checkbox is being assumed to be one character long and to start two characters back from the tail. That's not true if the checkbox is totally empty, and it's also not true if the checkbox were ever more than one character.

Instead of treating `head` as the `[` and `tail` as the `]`, you could define them as:
```js
const head = lineStart + checkBoxStart + 1
const tail = head + checkBoxStr.length - 2
```
Then you would be able to:
```js
const checkBoxInside = doc.getText([head, tail])
```
Then you could decide to delete based on the length of that string.

apps/tangent-electron/src/app/model/commands/ToDoToggleCommand.ts.ts
```diff
@@ -0,0 +1,116 @@
...
+
+	canExecute(context?: ToDoToggleCommandContext): boolean {
+		return this.getTargets(context) != null
+	}
+
+	execute(context?: ToDoToggleCommandContext): void {
+		const targets = this.getTargets(context)
+		if (!targets) return
+		const { editor, selection } = targets
+		const mark = this.workspace?.settings?.defaultTodoCompleteChar || 'x'
```
taylorhadden: You need to get the actual value of the setting object here:
```js
const mark = this.workspace?.settings?.defaultTodoCompleteChar?.value ?? 'x'
```

apps/tangent-electron/src/app/model/commands/ToDoToggleCommand.ts.ts
```diff
@@ -0,0 +1,116 @@
...
+		toggleToDo(editor, selection, 'x')
+	}
+
+	getLabel(context?: ToDoToggleCommandContext) {
+		return "Label"
```
taylorhadden: This and the tooltip should be filled out, but you knew that :)

apps/tangent-electron/src/app/model/commands/ToggleCheckboxCommand.ts.ts
```diff
@@ -0,0 +1,116 @@
...
+		let selection = context?.selection
+		let view = context?.view
+		
+		if (!view) {
+			view = getNoteView(this.workspace.viewState.tangent)
+			if (!view || !(view.editor instanceof MarkdownEditor)) return null
+		}
+
+		if (!editor || !selection) {
+			editor = editor || view.editor
```
taylorhadden, line 40: When doing value coalescing like this, using `??` is a better default to reach for. It allows non-null/undefined values that still evaluate to "falsey" like `0` and `''` to be treated as real values. `??` will take the first non-null, non-undefined value, and that's often what's wanted.

taylorhadden (changes requested):
Not sure exactly what's going on with the error you're seeing. The chromium developer console should log the error.

Another note: there should be tests for this toggle function made in `MarkdownEditor.test.ts`. There's already a `describe('List Handling') -> describe('Checkboxes')` section.

apps/tangent-electron/src/app/model/commands/ToggleCheckboxCommand.ts.ts
kenekoba: I've stolen this code from `apps\tangent-electron\src\app\model\commands\NoteFormattingCommands.ts`

kenekoba:
> The chromium developer console should log the error.

it does not print any useful information. somehow fixes the issue
```js
change.select([head, tail])
```

apps/tangent-electron/src/app/model/commands/ToDoToggleCommand.ts.ts
kenekoba: do you mean this?

1. if the selected line is a check box? toggle
2. if not, turn line into a empty checkbox

kenekoba: I did that but I'm not sure what to do when there are mix of checkboxes and non-checkboxes

kenekoba: Taylor I don't understand what's going on

it works on 
- single line
- multi-line `all align space`
- multi-line `all align empty`

but it does not work with "multi-line `mix space/empty`" or `mix style`
and I really could not figure out why 😥
it somehow sucks but the logs do not show anything

here's my test. **select multiple lines and then execute the command** . 

```md
all align space
- [ ] 1
- [ ] 2
- [ ] 3
- [ ] 4
- [ ] 5
- [ ] 6

all align empty
- [] 1
- [] 2
- [] 3
- [] 4
- [] 5
- [] 6

mix space/empty
- [] 1
- [] 2
- [ ] 4
- [ ] 5
- [] 6

mix style
1
- [] 2
- [ ] 3
- 4
5
- [ ] 6
```

here's the result
{EDA3E05C-17D5-4E48-9D42-E37C7ADBD463}


can you please check?

kenekoba: do I have to use `Delta` or something? it's not documented

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -474,6 +474,71 @@ export function toggleLineComment(editor: MarkdownEditor, event?: ShortcutEvent)
...
 
+export function toggleCheckbox(editor: Editor, selection: EditorRange, mark: string, forceTurnToCheckbox = true) {
+	const EMPTY_CHECKBOX = "[ ] "
+	const EMPTY_LIST_CHECKBOX = "- " + EMPTY_CHECKBOX
+	let addedCharactersCount = 0
+
+	const { doc, change } = editor
+	const [selectionStart, selectionEnd] = normalizeRange(selection)
+
+	// XXX use for loop version or while loop one
```
taylorhadden: For loop makes sense to me.

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -474,6 +474,71 @@ export function toggleLineComment(editor: MarkdownEditor, event?: ShortcutEvent)
...
+			else if (forceTurnToCheckbox) { // if line was not checkbox and not empty, make it a checkbox
+				const firstNonSpaceIndex = line.length - line.trimStart().length
+				change.insert(lineStart + firstNonSpaceIndex, EMPTY_LIST_CHECKBOX)
+				addedCharactersCount += EMPTY_LIST_CHECKBOX.length
+			}
+		}
+
+	}
+
+	change.select([selectionStart, selectionEnd + addedCharactersCount])
```
taylorhadden: If you execute this with selection on a single line, the start of your selection will end up offset. You want to adjust `selectionStart` by the number of characters added before it.

apps/tangent-electron/src/app/model/commands/ToggleCheckboxCommand.ts.ts
```diff
@@ -0,0 +1,76 @@
...
+	execute(context?: ToggleCheckboxCommandContext): void {
+		const targets = this.getTargets(context)
+		if (!targets) return
+		const { editor, selection } = targets
+		const mark = this.workspace?.settings?.defaultTodoCompleteChar.value
+		toggleCheckbox(editor, selection, mark)
+	}
+
+	getLabel(context?: ToggleCheckboxCommandContext) {
+		return "Switch Checkbox State"
```
taylorhadden: The rest of the commands use "Toggle" as a label.

apps/tangent-electron/src/app/model/commands/ToggleCheckboxCommand.ts.ts
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	getTooltip(context?: ToggleCheckboxCommandContext) {
+		return "Switches checkbox state on <-> off"
```
taylorhadden: I think "Switches checkbox state between on and off" is a little clearer. I want to avoid this kind of glyphs in tooltips. They generally don't look good.

taylorhadden (changes requested):
This is looking good! Just minor changes.

I would love to see new tests though.

kenekoba:
> I would love to see new tests though.

sure I'd add tests once I could get it working

kenekoba:
Taylor do you have any idea about this? https://github.com/suchnsuch/Tangent/pull/430#discussion_r3846257357

apps/tangent-electron/src/app/model/commands/ToDoToggleCommand.ts.ts
kenekoba: should I use `Delta`? what it's used for?

taylorhadden: Okay, this is strange. This may well be a situation where the `TextChange` mechanism is just failing to correctly adjust for deletions and insertions. The strangest part to me is that number 6 succeeds.

I would try (in order):
1. Switching the order of the inserts & deletes.
2. Manually tracking inserted & removed characters and using that to adjust insertions. It seems that insertions next to deletions are the things having trouble.
3. Giving up and writing your own `Delta`. I've ended up doing this _a lot_ for complex, multi-line transformations. See `setLinePrefix()` in `editorActions.ts` for a simple example (which I wish I documented better).

Honestly, this is the perfect thing to add to some automated tests. It can really speed up working through the problem. I use the "Vitest" plugin with VSCode. It makes debugging and stepping through functions trivial. Highly recommended when working through weird stuff like this.

kenekoba: wow swapping `delete` and `insert` fixed the issue! thanks!!!!!

kenekoba:
Taylor it's ready! I've added tests

taylorhadden (approved):

taylorhadden:
Awesome, thank you!

Looking at the tests, I may adjust the behavior so that the target checkbox action is consistent across all targets. For example, given:
```md
- A
- [ ] B
- [x] C
- D
```
…it might be better for selecting all and toggling on to only make "A" and "D" turn into checkboxes.

For another example, it might be better if selecting all of this:
```md
- [ ] A
- [x] B
- [ ] C
- [x] D
```
and toggling checked off "A" and "C" turned on.

And then finally, selecting and toggling this:
```md
- [x] A
- [x] B
- [x] C
- [x] D
```
would turn them all off.

This would let a user select a series of any lines and execute the command multiple times to move the whole selection towards being completed and uniform. This seems better than just inverting every line in the selection, as you can be a lot less precise in the lines you select.

PR #431, opened 2026-08-22
fix `findWordAround` for non-Latin characters

### LINKED ISSUE #429: `findWordAround` does not capture non-Latin words
Hi, 

commands like `bold` and `italic`, the function `findWordAround` is used to selected a word when the selection is none (i.e. head == tail)

it's good to support non-Latin words too.

https://github.com/suchnsuch/Tangent/blob/66fcc1405e12f76a168ee5bba459003dc8672bcc/apps/tangent-electron/src/common/stringUtils.ts#L50-L58

kenekoba:
fixes https://github.com/suchnsuch/Tangent/issues/429

### DIFF
--- a/apps/tangent-electron/src/common/stringUtils.ts
+++ b/apps/tangent-electron/src/common/stringUtils.ts
@@ -48,7 +48,7 @@ export function searchWhileCharacterMatches(text: string, match: RegExp, start: 
 }
 
 export function findWordAround(text: string, position: number) {
-	const match = /[\w\d_-]/
+	const match = /[\S\d_-]/
 	const behindSearch = searchWhileCharacterMatches(text, match, position, -1)
 	const wordStart = behindSearch.foundAnyMatches ? behindSearch.index : position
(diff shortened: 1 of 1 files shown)


apps/tangent-electron/src/common/stringUtils.ts
```diff
@@ -48,7 +48,7 @@ export function searchWhileCharacterMatches(text: string, match: RegExp, start:
 }
 
 export function findWordAround(text: string, position: number) {
-	const match = /[\w\d_-]/
+	const match = /[\S\d_-]/
```
taylorhadden, line 51: I'm wondering if it's better to make this a blocklist `/[^]/` instead. e.g. `/[^\s ()\[\]]/` etc.

The best thing to do would be to expand the tests in `stringUtils.test.ts` for non-Latin characters and the other things that the function should do.

If you prefer, I can take a look at this. There are things this function does right now that are not captured by tests, and there are things that it probably _should_ do that it doesn't.

taylorhadden (changes requested):

PR #442, opened 2026-08-30
force textarea only to grow vertically

hamidb80:
it prevents issues like this to break the UI:

image


taylorhadden (approved):
Oh that's cool. I did not know you could do that. 🔥🔥🔥

PR #445, opened 2026-08-31
Show the new-note destination in command tooltips

snowyukitty:
Resolves the Create New Note tooltip destination from explicit, rule, path, or selected-folder context without invoking creation preparation, while retaining custom rule descriptions. Adds focused regression coverage for default, nested, invalid, transition, purity, repeated-read, Unicode, and full-path behavior.

Tests, from `apps/tangent-electron`:

- `npx vitest run src/app/model/commands/CreateNewFile.test.ts` — 13 passed
- `npx tsc --noEmit -p tsconfig.json` — no output, exit 0

`npx vitest run --project tangent-electron-unit-core --project tangent-electron-unit-web` is unchanged by this branch: run once at its head and once with the two changed files checked out at the base, both report `Test Files 9 failed | 28 passed | 1 skipped (38)` and `Tests 3 failed | 320 passed | 12 skipped (335)`. That is not a clean run and I am not claiming one — the 3 failing tests are `paths.test.ts` `resolve()` cases, and 8 of the 9 failing files never collect, because `@such-n-such/tangent-query-parser/lib/types` is not built in my checkout.

### DIFF
--- a/apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
+++ b/apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
@@ -3,16 +3,31 @@ import CreateNewFileCommand, { type CreateNewFileCommandContext } from './Create
 import type { Workspace } from '..'
 import IndexTreeStore from 'common/indexing/IndexTreeStore'
 import { knownExtensions } from 'common/fileExtensions'
+import type { TreeNode } from 'common/trees'
 
 describe('Extension auto inclusion', () => {
 
+	const unicodeFolder: TreeNode = {
+		name: '研究',
+		path: 'some/root/Ideas/研究',
+		depth: 3,
+		fileType: 'folder',
+		children: []
+	}
+	const ideasFolder: TreeNode = {
+		name: 'Ideas',
+		path: 'some/root/Ideas',
+		depth: 2,
+		fileType: 'folder',
+		children: [unicodeFolder]
+	}
 	const directoryStore = new IndexTreeStore({
 		files: {
 			name: 'root',
 			path: 'some/root',
 			depth: 1,
 			fileType: 'folder',
-			children: []
+			children: [ideasFolder]
 		},
 		tags: {
 			path: '',
@@ -21,17 +36,20 @@ describe('Extension auto inclusion', () => {
 			fileType: ''
 		}
 	})
+	const effects = { navigation: 0, modal: 0 }
 
-	const workspace = {
+	const workspace: Workspace = Object.assign(Object.create(null), {
 		directoryStore,
+		navigateTo: () => effects.navigation++,
 		viewState: {
+			modal: { push: () => effects.modal++ },
 			directoryView: {
 				selection: {
 					value: []
 				}
 			}
 		}
-	} as Workspace
+	})
 
 	const command = new CreateNewFileCommand(workspace)
 	// Expose the private function type-safely
@@ -108,5 +126,51 @@ describe('Extension auto inclusion', () => {
 			extension: '.md',
 			creationMode: undefined
 		})
+	})
+
+	it('Describes the default, selected, nested, and Unicode destinations', () => {
+		const selection = workspace.viewState.directoryView.selection
+		selection.value = []
+		expect(command.getTooltip()).toContain('Destination: Workspace Root')
+
+		selection.value = [unicodeFolder]
+		expect(command.getTooltip()).toContain('Destination: Ideas/研究')
+	})
+
+	it('Does not claim invalid destinations from reserved or invalid inputs', () => {
+		workspace.viewState.directoryView.selection.value = [{
+			name: 'private', path: 'some/root/.tangent/private', depth: 3, fileType: 'folder'
+		}]
+		expect(command.getTooltip()).toContain('Destination: Workspace Root')
+		expect(command.getTooltip({ relativePath: 'CON/New Note.md' })).toBe('Creates a new note')
+	})
+
+	it('Updates the destination for path and explicit-folder contexts', () => {
+		workspace.viewState.directoryView.selection.value = [unicodeFolder]
+		expect(command.getTooltip({ relativePath: 'Projects/Long Term/New Note.md' }))
+			.toContain('Destination: Projects/Long Term')
+		expect(command.getTooltip({ folder: ideasFolder }))
+			.toContain('Destination: Ideas')
+	})
+
+	it('Preserves rule descriptions and resolves rule destination precedence', () => {
+		const context: CreateNewFileCommandContext = {
+			folder: ideasFolder,
+			path: 'some/root/Other/New Note.md',
+			rule: {
+				name: 'Journal', nameTemplate: 'Daily', folder: 'Journal', contentTemplate: '',
+				mode: 'create', description: 'A custom journal description'
+			}
+		}
+		const before = structuredClone(context)
+		const beforeSelection = workspace.viewState.directoryView.selection.value
+		const beforeEffects = { ...effects }
+		const tooltip = command.getTooltip(context)
+		expect(tooltip).toContain('A custom journal description')
+		expect(tooltip).toContain('Destination: Ideas')
+		expect(command.getTooltip(context)).toBe(tooltip)
+		expect(context).toEqual(before)
+		expect(workspace.viewState.directoryView.selection.value).toBe(beforeSelection)
+		expect(effects).toEqual(beforeEffects)
 	})
 })
--- a/apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
+++ b/apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
@@ -4,30 +4,33 @@ import type { Workspace } from '..'
 import IndexTreeStore from 'common/indexing/IndexTreeStore'
 import { knownExtensions } from 'common/fileExtensions'
 import type { TreeNode } from 'common/trees'
+import type { CreationRuleDefinition } from 'common/settings/CreationRule'
 
 describe('Extension auto inclusion', () => {
 
-	const unicodeFolder: TreeNode = {
-		name: '研究',
-		path: 'some/root/Ideas/研究',
-		depth: 3,
-		fileType: 'folder',
-		children: []
-	}
-	const ideasFolder: TreeNode = {
-		name: 'Ideas',
-		path: 'some/root/Ideas',
-		depth: 2,
-		fileType: 'folder',
-		children: [unicodeFolder]
-	}
 	const directoryStore = new IndexTreeStore({
 		files: {
 			name: 'root',
 			path: 'some/root',
 			depth: 1,
 			fileType: 'folder',
-			children: [ideasFolder]
+			children: [
+				{
+					name: 'Ideas',
+					path: 'some/root/Ideas',
+					depth: 2,
+					fileType: 'folder',
+					children: [
+						{
+							name: '研究',
+							path: 'some/root/Ideas/研究',
+							depth: 3,
+							fileType: 'folder',
+							children: []
+						}
+					]
+				}
+			]
 		},
 		tags: {
 			path: '',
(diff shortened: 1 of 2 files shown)


apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -3,16 +3,31 @@ import CreateNewFileCommand, { type CreateNewFileCommandContext } from './Create
 import type { Workspace } from '..'
 import IndexTreeStore from 'common/indexing/IndexTreeStore'
 import { knownExtensions } from 'common/fileExtensions'
+import type { TreeNode } from 'common/trees'
 
 describe('Extension auto inclusion', () => {
 
+	const unicodeFolder: TreeNode = {
```
taylorhadden: I would prefer these to be declared inline so that the tree structure is clear. The `const` variables could then be initialized from their position in the tree, e.g. `const ideasFolder = files.children[0]`

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -21,17 +36,20 @@ describe('Extension auto inclusion', () => {
 			fileType: ''
 		}
 	})
+	const effects = { navigation: 0, modal: 0 }
 
-	const workspace = {
+	const workspace: Workspace = Object.assign(Object.create(null), {
```
taylorhadden: Why are you doing this like this?

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -375,12 +375,63 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
 		return 'Create New Note'
 	}
 
-	getTooltip(context: CreateNewFileCommandContext) {
+	private resolveTooltipFolder(context?: CreateNewFileCommandContext) {
```
taylorhadden: This just duplicates the logic for determining the folder. Can't the `resolveContext()` function be used instead?

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -375,12 +375,63 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
...
 		}
-		return 'Creates a new note'
+		else {
+			description = 'Creates a new note'
+		}
+
+		const folder = this.resolveTooltipFolder(context)
+		if (folder === undefined || folder === false) return description
+		const destination = !folder || folder === '.' ? 'Workspace Root' : folder
+		return `${description}\nDestination: ${destination}`
```
taylorhadden: When there isn't a custom description, it will probably read better as: "Creates a new <name> in <My/Folder>." or "Creates a new <name> in the root of the workspace."

Appending `\nDestination:…` after a full description makes sense to me.

taylorhadden (changes requested):
This really pushes me in the direction that commands want to be able to author full complex tooltips (or perhaps just some HTML). That would let there be full formatting for what's being provided here. That doesn't need to be part of this spec.

snowyukitty:
Pushed `9e41af5`, with current main merged in. Against main the branch is 2 files, +158/-19.

**`resolveContext()` instead of the duplicate.** `resolveTooltipFolder()` is gone. `resolveContext()` now takes `{ interactive: false }`, which is what a tooltip needs from it: it never pushes `CreateFileDialog`, it returns `undefined` rather than throwing on a path `validatePath()` rejects or a folder it cannot make relative, and it skips a selected item with no parent instead of reading `.depth` off it. `execute()` passes no options, so all three keep today's behaviour — the parentless guard is `!interactive && !folder` specifically so file creation still takes the same path through that loop as it does today.

Reusing the real resolver fixed one thing the duplicate had quietly been getting wrong, and surfaced one case that has no honest answer:

- A rule whose `nameTemplate` defines a folder used to show only the rule's own folder. It now resolves the template too, because the prompted slot is filled with the default name: `Meetings/%name%` under a `Notes` rule folder shows `Notes/Meetings`.
- When the prompted slot comes *before* a separator (`%name%/Entry`), the name the user types is what picks the folder, so there is nothing true to show yet. The tooltip drops the destination rather than naming a folder the file will not go in.

**Wording.** With no custom description it now reads `Creates a new note in Ideas/研究.` or `Creates a new note in the root of the workspace.`, and a rule without its own description reads `Creates a new Journal in Journal.`. A custom description is kept verbatim and gains the `Destination:` line.

**Tests.** The tree is declared inline and `ideasFolder` / `unicodeFolder` are taken from their positions in it. The workspace stub is an ordinary object literal again — the `Object.create(null)` was me working around the `as Workspace` assertion once `modal.push` was added to `viewState`, which `as unknown as Workspace` handles without the misdirection. Sorry for the noise. The selection value is typed `TreeNode[]` so the tests can reassign it.

The new cases lock both sides of the flag: hovering a prompting rule must not increment the modal counter, and the same rule through `resolveContext()` with the default must.

**Validation at `9e41af5`.** `tsc --noEmit -p apps/tangent-electron` is clean. `npx vitest run` gives 17 failed, 532 passed, 13 skipped; a detached checkout of current main gives 17 failed, 525 passed, 13 skipped. Identical failure set — the `paths.test.ts` `resolve()` cases and the `workspace-querying` integration tests, both of which fail on this Windows machine with the branch absent.

On commands authoring richer tooltips: agreed that is the better shape, and I have kept this to the plain-string interface so it does not get in the way.

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -170,6 +188,9 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
 
 				// When looking at a folder, create items within the folder
 				let folder = item.fileType === 'folder' ? item : directoryStore.getParent(item)
+				// A parentless item cannot contribute a folder. Only the side-effect-free
+				// path skips it; execute() keeps whatever it does with one today.
```
taylorhadden: This is very clearly a comment written by claude. It's also not helpful. When is `folder` nullish here?

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -109,4 +133,87 @@ describe('Extension auto inclusion', () => {
...
 		})
 	})
+
+	it('Describes the default, selected, nested, and Unicode destinations', () => {
+		const selection = workspace.viewState.directoryView.selection
+		selection.value = []
+		expect(command.getTooltip()).toBe('Creates a new note in the root of the workspace.')
+
+		selection.value = [unicodeFolder]
+		expect(command.getTooltip()).toBe('Creates a new note in Ideas/研究.')
```
taylorhadden: Target folders should be contained in "Quotes"

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -109,4 +133,87 @@ describe('Extension auto inclusion', () => {
...
+	})
+
+	it('Does not claim invalid destinations from reserved or invalid inputs', () => {
+		workspace.viewState.directoryView.selection.value = [{
+			name: 'private', path: 'some/root/.tangent/private', depth: 3, fileType: 'folder'
+		}]
+		expect(command.getTooltip()).toBe('Creates a new note in the root of the workspace.')
+		expect(command.getTooltip({ relativePath: 'CON/New Note.md' })).toBe('Creates a new note')
```
taylorhadden: The tooltip should always end with a period.

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -109,4 +133,87 @@ describe('Extension auto inclusion', () => {
...
+			rule: {
+				name: 'Journal', nameTemplate: 'Daily', folder: 'Journal', contentTemplate: '',
+				mode: 'create', description: 'A custom journal description'
+			}
+		}
+		const before = structuredClone(context)
+		const beforeSelection = workspace.viewState.directoryView.selection.value
+		const beforeEffects = { ...effects }
+		const tooltip = command.getTooltip(context)
+		expect(tooltip).toContain('A custom journal description')
```
taylorhadden: Can we assert against the full tooltip here?

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -109,4 +133,87 @@ describe('Extension auto inclusion', () => {
...
+
+	it('Claims no destination when the typed name would choose the folder', () => {
+		workspace.viewState.directoryView.selection.value = []
+		const beforeEffects = { ...effects }
+		expect(command.getTooltip({
+			rule: {
+				name: 'Entry', nameTemplate: '%name%/Entry', folder: 'Notes',
+				contentTemplate: '', mode: 'create', description: ''
+			}
+		})).toBe('Creates a new Entry')
```
taylorhadden: Isn't calling out the parent folder valid here?

taylorhadden (changes requested):

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
taylorhadden: To be clear: I don't really care whether you use Claude or not, but any comments in the code need to make sense. Claude is very good at making non-sensical comments.

snowyukitty:
Pushed `50f1437`. Against `main` the branch is 2 files, +169/-19.

**When `folder` is nullish.** Only when the selected item is not a folder and `directoryStore.getParent()` misses on it — a selected root is a folder, so it takes the other branch and `folder` is the item itself. `getParent()` is a `parentLookup.get()`, so a miss is a node the store no longer holds a parent entry for, which in practice means one removed since it was selected; `directoryStore.test.ts` asserts `getParent()` is `undefined` after a `remove()`. The next line reads `folder.depth`.

The comment now names that condition, and only that one. I scoped the guard itself to `!interactive` because I did not want a tooltip change to alter what the creation path does with such a selection; if you would rather it covered both paths, say so and I will drop the `!interactive &&`.

**Quotes.** Destinations are quoted now: `Creates a new note in "Ideas/研究".` That matches `getLabel()` in `DeleteNode.ts` and `DuplicateNode.ts`, which both render `` `Delete "${target.name}"` ``-style names. "the root of the workspace" is prose rather than a folder name, so it stays unquoted.

**The period.** Every tooltip this builds ends in one, including the no-destination fallback, `Creates a new note.`, and the description form, `…\nDestination: "Ideas".` The one string left alone is a rule's own `description` when it is returned by itself, because no destination could be resolved — that text is the author's, so I did not append to it. Say the word if you want a period forced there too.

**The full tooltip.** The two `toContain` assertions are now one `toBe` against `'A custom journal description\nDestination: "Ideas".'`.

**The parent folder.** You are right, and I was throwing it away. For `%name%/Entry` in folder `Notes`, everything before the token is fixed before anything is typed, so `Notes` is known even though the leaf folder is not: `Creates a new Entry in "Notes".` The preview now drops only the part of the template after the token instead of abandoning the destination entirely, so `Archive/%name%/Entry` resolves to `Notes/Archive`. Both are asserted.

Verification, all local: `vitest run src/app/model/commands/CreateNewFile.test.ts` — 13 passed. `tsc --noEmit -p tsconfig.json` from `apps/tangent-electron` — no output, exit 0. For the wider suites I ran `tangent-electron-unit-core` and `tangent-electron-unit-web` twice, once at `50f1437` and once with the two files checked out at `9e41af5`; both runs print `Test Files 9 failed | 28 passed | 1 skipped (38)` and `Tests 3 failed | 320 passed | 12 skipped (335)`. The 3 failing tests are `paths.test.ts` `resolve()` cases, and 8 of the 9 failing files never collect, because `@such-n-such/tangent-query-parser/lib/types` is not built in my checkout. So that is not a clean run and I am not claiming one — only that reverting this change does not alter it.

PR #446, opened 2026-09-02
add preview for copy path commands as `tooltip`

hamidb80:
Hi, sometimes I wanna abuse the `copy path` command "to see path of current note/file". based on [Taylor's advice](https://github.com/suchnsuch/Tangent/pull/388#discussion_r3905450457) the tooltip should contain useful information, in this context I think seeing the path is useful

image

### DIFF
--- a/apps/tangent-electron/src/app/model/commands/CopyPath.ts
+++ b/apps/tangent-electron/src/app/model/commands/CopyPath.ts
@@ -6,8 +6,8 @@ interface CopyPathCommandContext extends CommandContext {
 	node: TreeNode
 }
 
-function copyPathTooltip(path: string): string {
-	return `Copies "${path}" to the clipboard`
+function copyTooltip(what: string): string {
+	return `Copies "${what}" to the clipboard`
 }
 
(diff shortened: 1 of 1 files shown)


taylorhadden (changes requested):
This looks great! My further advice would be that the tooltip should ideally read as a sentence. So, in this case: "Copies "<path>" to the clipboard."

apps/tangent-electron/src/app/model/commands/CopyPath.ts
```diff
@@ -6,6 +6,10 @@ interface CopyPathCommandContext extends CommandContext {
 	node: TreeNode
 }
 
+function copyTooltip(what: string): string {
+	return `Copies "${what}" to the clipboard`
```
taylorhadden: Final nit! Since tooltips should read like sentences, this should end with a period.

taylorhadden (changes requested):

taylorhadden (approved):

taylorhadden:
Thank you!

Issue #452, opened 2026-09-11
Using "Creation Rules" to create a note in relative path


hamidb80:
Hi,

I wanted my new notes to be named `temp` instead of `New Note`, and I also didn't want them to be created incrementally (e.g., `New Note 1`, `New Note 2`, etc.).

So I registered a new "Creation Rule" like this:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/60a8537b-127b-4645-bf40-08824414717b" />

But it fails to create the note in the local folder:

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/6fa50941-2441-4bfe-979a-9cb178372eea" />

```
CreateNewFile.ts:141 Uncaught Error: Could not create file. "./temp.md" is invalid and could not be made valid.
    at $3.resolveContext (CreateNewFile.ts:141:11)
    at $3.execute (CreateNewFile.ts:305:23)
    at _ (CommandPalette.svelte:370:11)
    at HTMLElement.g (ScrollingItemList.svelte:70:37)
```

----

* Tangent Notes v0.13.0-alpha
* OS: Windows 10

taylorhadden:
The parent folder field expects an exact folder at the moment. This use case is not currently supported.

PRs to fix that would be welcome!

PR #454, opened 2026-09-13
add relative path support for creation rule

hamidb80:
addresses https://github.com/suchnsuch/Tangent/issues/452


apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -117,7 +117,21 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
 			}
 
 			// Combining into a relative path allows rule name templates to define folders
-			relativePath = paths.join(folderPath || rule.folder, name + '.md')
+			let relativeDirPath = folderPath || rule.folder
+			if (relativeDirPath.startsWith('.')) {
+				const workspaceDir = this.workspace.viewState.directoryView.root.path
+				const viewState = this.workspace.viewState.tangent.getCurrentViewState()
+				const absoluteParentDir = viewState 
+								? (viewState.node.fileType == 'folder')
```
taylorhadden: Nit: this is a very large amount of indentation here. I recommend a single additional tab from the current scope.

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -117,7 +117,21 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
 			}
 
 			// Combining into a relative path allows rule name templates to define folders
-			relativePath = paths.join(folderPath || rule.folder, name + '.md')
+			let relativeDirPath = folderPath || rule.folder
+			if (relativeDirPath.startsWith('.')) {
+				const workspaceDir = this.workspace.viewState.directoryView.root.path
+				const viewState = this.workspace.viewState.tangent.getCurrentViewState()
```
taylorhadden: You could also do `...getCurrentViewState()?.node` as you only ever reference `node` in the view state.

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -117,7 +117,21 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
...
+				const absoluteParentDir = viewState 
+								? (viewState.node.fileType == 'folder') 
+									? viewState.node.path                // in folder
+									: paths.dirname(viewState.node.path) // in file
+								:  workspaceDir                          // at root level
+				const absoluteParentPath = paths.join(absoluteParentDir, relativeDirPath)
+				relativeDirPath = paths
+								.resolve(absoluteParentPath) // resolve . or ..
+				    			.substring(workspaceDir.length) // make path relative to the workspace
```
taylorhadden: While this should work, I recommend using `workspace.directoryStore.pathToRelativePath()`. Alternatively, since you know the root already, you can use `paths.getChildPath()`.

taylorhadden (changes requested):
Looks pretty solid overall. None of my comments are big issues.

I would love to see some tests added to `CreateNewFile.test.ts`!

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
hamidb80: no I've used `viewState.node` 3 times in the code, removing the `viewState` makes it messy

hamidb80:
Hi @taylorhadden . sorry for interruption.

seems like `workspace.viewState.tangent` is not available at test time!?!. what should I do?


```
TypeError: Cannot read properties of undefined (reading 'getCurrentViewState')
 ❯ CreateNewFileCommand.resolveContext src/app/model/commands/CreateNewFile.ts:122:50
    122|     const node = this.workspace.viewState.tangent.getCurrentViewState(…
       |                                                  ^
```

taylorhadden:
Looking at this again, does this actually handle the case described in #452? To resolve a path with a relative section, we need three parts: the base, the relative folder, and the name section. It seems like the base is replacing the relative section.

hamidb80:
> does this actually handle the case described in #452? 

Yes I've tested it manually. It works in these cases:
1. Root (no file opened)
2. In a directory
3. In a file

The test path was `./` and the file name was `temp`

hamidb80:
It may need more tests but I first need to get around https://github.com/suchnsuch/Tangent/pull/454#issuecomment-5691108905

taylorhadden:
Sorry, I missed that. At the top of the file for `CreatenewFile.test.ts`, there is a `const workspace` definition that contains a bunch of dummy mock objects. You can either choose to implement a mock that gets you what you need for the test, or you can configure the code to handle a null `.tangent` field.

What happens if the test path is something other than `./`? For instance `../MyFolder` or `./OtherFolder`? We can't allow relative file paths to escape out of the workspace.

hamidb80:
> What happens if the test path is something other than `./`? For instance `../MyFolder` or `./OtherFolder`? We can't allow relative file paths to escape out of the workspace.

I've added some tests. the path would not jump out of workspace

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -3,39 +3,46 @@ import CreateNewFileCommand, { type CreateNewFileCommandContext } from './Create
...
-
-	const workspace = {
-		directoryStore,
-		viewState: {
-			directoryView: {
-				selection: {
-					value: []
+	// Expose the private function type-safely
+	function resolveContext(context: CreateNewFileCommandContext, node?: TreeNode) {
+		const directoryStore = new IndexTreeStore({
```
taylorhadden, line 11: FWIW, this `directoryStore` and `workspace` is never changed and can be a const value outside of the function so that it doesn't get reconstructed on every use.

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -109,4 +116,85 @@ describe('Extension auto inclusion', () => {
...
+				mode: 'createOrOpen',
+				contentTemplate: undefined,
+				description: undefined
+			},
+			extension: 'default-md'
+		}, {
+				path: '/some/root/note.md',
+				name: 'test note',
+				fileType: 'note',
+				createShallowCopy(){ return {} as TreeNode }
```
taylorhadden, line 135: Is this stub actually necessary?

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
```diff
@@ -109,4 +116,132 @@ describe('Extension auto inclusion', () => {
 			creationMode: undefined
 		})
 	})
+
+	it('relative path in note: ./', () => {
```
taylorhadden, line 120: Minor style note: the convention is to write the name of the test as if it was a descriptive sentence starting with "it". For instance, here might be "(it) Resolves paths adjacent to a note" or something. The name of the test should broadly assert what the expectation is.

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -117,7 +117,18 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
...
+			let relativeDirPath = folderPath || rule.folder
+			if (relativeDirPath.startsWith('.')) {
+				const node = this.workspace.viewState.tangent.getCurrentViewState()?.node
+				const relativeNodePath = node && 
+					this.workspace.directoryStore.pathToRelativePath(
+						(node.fileType == 'folder') 
+							? node.path                // in folder
+							: paths.dirname(node.path) // in file
+					) || '' // at root level
+				relativeDirPath = paths.resolve(relativeNodePath + '/' + relativeDirPath)
```
taylorhadden, line 129: Don't manually insert path separators. Always use `paths.join()`.

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
```diff
@@ -117,7 +117,18 @@ export default class CreateNewFileCommand extends WorkspaceCommand {
...
+			}
+			relativePath = paths.join(relativeDirPath, name + '.md')
```
taylorhadden, line 131: If the named template also includes a relative path, this will break here. Perhaps worth fixing with a `resolve()`?

taylorhadden (changes requested):
Looking good! Just some relatively minor notes besides the manual slash inclusion.

apps/tangent-electron/src/app/model/commands/CreateNewFile.test.ts
hamidb80: I'll remove `createShallowCopy`

apps/tangent-electron/src/app/model/commands/CreateNewFile.ts
hamidb80: I didn't change this line. the behavior is as before.

maybe a test could reveal the bug?

taylorhadden: Sure, but since relative path support is being added for the folder, it makes sense to do so from the name template as well, no?

Issue #438, opened 2026-08-26
highlight toggle should be consistent with each other


kenekoba:
for example applying `blue` highlight should replace `yellow`

taylorhadden:
Is this something you want to tackle after #430?

hamidb80:
I'm working on a uniform emoji predicate in https://github.com/suchnsuch/Tangent/pull/444 . 
it's rather strange since the byte length of white circle and colorful circles differ:

```js
const getByteLength = (char) => new TextEncoder().encode(char).length 
console.log(getByteLength('⚪')) // 3
console.log(getByteLength('🟡')) // 4
```

taylorhadden:
Yeah, emoji are _weird_.

kenekoba:
> Is this something you want to tackle after [#430](https://github.com/suchnsuch/Tangent/pull/430)?

Hi, I don't know where to start. the existing `toggleInlineFormat` seems a little bit complicated. should I rewrite the logic?

taylorhadden:
@kenekoba I don't think so. `toggleInlineFormat()` is lower level than whatever we're looking for here I think. @hamidb80, if you think you're making progress here, that's fine by me!

PR #456, opened 2026-09-14
general highlight

kenekoba:
addresses. https://github.com/suchnsuch/Tangent/issues/438

somehow the video cannot be played in Firefox, only works on Chrome-based browsers :/

## Demo

https://github.com/user-attachments/assets/a81de839-9deb-4002-b23e-73f3759d8d6c

## Config
I change the highlight by `shortcuts`:

{F4321381-B21A-45EF-94A9-636ED44DE9A1}

----


kenekoba:
I think the next step would be to add tests, right?

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+		if (range) {
+			ranges.push(range)
+		}
+	}
+
+	console.log(ranges.map(([a, b]) => editor.getText([a, b])))
+	let previousHighlight = ''
+	const change = editor.change
+
+	const origAt = at
```
taylorhadden: This is a style thing, but I much prefer variable names without contractions in them.

Also, it looks like `at` an `to` aren't written to afterwards, so this two particular variables may not be needed?

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+	if (willInsert) {
+		if (origAt === origTo) {
+			target = findWordAroundPositionInDocument(doc, origAt)
+		}
+		targetStart = target[0]
+		targetEnd = target[1]
+		newL = highlightToken.length
+	}
+
+	// ---- Single loop: delete then insert per range, back-to-front ----
```
taylorhadden: This is an unnecessary (or misplaced) comment.

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+		console.log('Toggle off')
+
+		// Process from last range to first so earlier coords stay valid
+		const orderedRanges = [...ranges].sort((a, b) => b[0] - a[0])
+
+		for (const range of orderedRanges) {
+			const [start, end] = range
+			const text = editor.getText(range)
+
+			const match = text.match(highlightEmojiMatch)
```
taylorhadden, line 202: What happens if this `match()` fails?

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+				const replacement = highlightToken + inner + highlightToken
+
+				change.delete([start, end])
+				change.insert(start, replacement)
+				affectedLineCount++
+			}
+		}
+	}
+	else if (willInsert) {
+		console.log('Toggle on')
```
taylorhadden: This should be removed

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+			prevL = previousHighlight.length
+
+			if (previousHighlight === highlightToken) {
+				// Pure delete for this range
+				change
+					.delete([end - prevL, end])
+					.delete([start, start + prevL])
+			}
+			else {
+				// Atomic replace: strip old markers, wrap with new ones
```
taylorhadden, line 213: This would mean that a selection that crosses two highlighted areas with a gap in the middle would only change the color of the highlighted areas and leave the gap untouched, right?

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+export function toggleInlineHighlightFormat(editor: Editor, selection: EditorRange, highlightToken: string) {
+	console.log(highlightToken)
+
+	const { doc } = editor
+	selection = normalizeRange(selection)
+	if (!selection) return
+	let [at, to] = selection
+
+	const predicate: AttributePredicate = attr => {
+		// console.log(attr)
```
taylorhadden, line 145: This log and the one below can be removed.

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+
+			change
+				.insert(s, highlightToken)
+				.insert(e, highlightToken)
+		}
+	}
+
+	// ---- Phase 3: Compute final selection ----
+	if (willInsert && previousHighlight !== highlightToken) {
+		const L = newL
```
taylorhadden: Again, I much prefer full variable names over contractions. I assume this refers to "length".

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+		const deltaPerSide = L - prevL
+
+		if (origAt === origTo && targetStart !== targetEnd && origAt === targetEnd) {
+			console.log('aa')
```
taylorhadden: Remove

apps/tangent-electron/src/app/views/editors/NoteEditor/editorActions.ts
```diff
@@ -131,6 +132,149 @@ export function toggleInlineFormat(editor: Editor, selection: EditorRange, forma
...
+			change.select(origAt + (L - prevL) + L)
+		}
+		else {
+			console.log('bb')
```
taylorhadden: Remove

taylorhadden (changes requested):
My big wonder is what happens if two (or more) non-adjacent highlighted spans are selected and the user presses a toggle command for a different highlight token. My naive assumption as a user would be that the two highlighted sections are converted to the new target, but also that it would be unified into a single highlighted span.

Also, please add some tests! They should go in `MarkdownEditor.test.ts`.

